PR #1, opened 2015-01-19
Remove the (pointless) dependency on libstd

Tobba:
By adding #![no_std].

This makes the library much more useful for low-level users.
It still needs the user to have an "std" module, but pretty much every no_std project already has a faux-std module around for this reason.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,9 @@
 
 //! A typesafe bitmask flag generator.
 
+#![no_std]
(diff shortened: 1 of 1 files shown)


Tobba:
Uh, this was poorly thought out. Give me a second.

FlaPer87:
@Tobba could you squash those 2 commits? Otherwise, it looks good to me

Tobba:
Squashed.

alexcrichton:
Thanks!

PR #3, opened 2015-02-04
added bitflags_values! helper, now possible to add values later.

Centril:
This PR adds a new helper macro `bitflags_values!`.
This was done so that you can:

``` .rust
 #[macro_use]
 extern crate bitflags;

 bitflags! { flags Flags: u32 }
 bitflags_values! {
     Flags {
         const FLAG_A       = 0b00000001,
         const FLAG_B       = 0b00000010,
         const FLAG_C       = 0b00000100,
         const FLAG_ABC     = FLAG_A.bits
                            | FLAG_B.bits
                            | FLAG_C.bits,
     }
 }

 fn main() {
     let e1 = FLAG_A | FLAG_C;
     let e2 = FLAG_B | FLAG_C;
     assert!((e1 | e2) == FLAG_ABC);   // union
     assert!((e1 & e2) == FLAG_C);     // intersection
     assert!((e1 - e2) == FLAG_A);     // set difference
     assert!(!e2 == FLAG_A);           // set complement
 }
```

Basically, it means that you don't have to create the values immediately.

The use case that necessitated this is to extend `bitflags!` with a new macro that allows you to create `bitflags` which start from `0x1` and shifts the value `1 bit` to the `left` for every flag making all flags _mutually exclusive_.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,67 @@
 #![no_std]
 #[cfg(test)] #[macro_use] extern crate std;
 
+/// The `bitflags_values` macro is a helper for the `bitflags!` macro.
+/// It creates the actual constants and the implementation for the `all` method.
+///
+/// It is used internally, but can be useful if you want to create macros
+/// extending `bitflags!`. For example, if you can't set all values from
+/// the get go, you can later use this.
+///
+/// ```{.rust}
+/// #[macro_use]
+/// extern crate bitflags;
+///
+/// bitflags! {
+///     flags Flags: u32
+/// }
+/// bitflags_values! {
+///     Flags {
+///         const FLAG_A       = 0b00000001,
+///         const FLAG_B       = 0b00000010,
+///         const FLAG_C       = 0b00000100,
+///         const FLAG_ABC     = FLAG_A.bits
+///                            | FLAG_B.bits
+///                            | FLAG_C.bits,
+///     }
+/// }
+///
+/// fn main() {
+///     let e1 = FLAG_A | FLAG_C;
+///     let e2 = FLAG_B | FLAG_C;
+///     assert!((e1 | e2) == FLAG_ABC);   // union
+///     assert!((e1 & e2) == FLAG_C);     // intersection
+///     assert!((e1 - e2) == FLAG_A);     // set difference
+///     assert!(!e2 == FLAG_A);           // set complement
+/// }
+/// ```
+///
+#[macro_export]
+macro_rules! bitflags_values {
+    ($BitFlags:ident {
+        $($(#[$Flag_attr:meta])* const $Flag:ident = $value:expr),+
+    }) => {
+        $($(#[$Flag_attr])* pub const $Flag: $BitFlags = $BitFlags { bits: $value };)+
+
+        impl $BitFlags {
+            /// Returns the set containing all flags.
+            #[inline]
+            pub fn all() -> $BitFlags {
+                $BitFlags { bits: $($value)|+ }
+            }
+        }
+    };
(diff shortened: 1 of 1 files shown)


brson:
I'm curious to see the macro you wrote using this as an example of how it will be used.

Centril:
Working on it here: https://github.com/Lumenix/rust-exbitflags
Not up to date, but the basic idea is there - it doesn't work yet :/
Probably a good idea to wait with merging until I can show that it actually works =)

alexcrichton:
This macro is basically already exposing the implementation detail that the structure generated is a struct with one `bits` field, so could that knowledge be leveraged instead? Using a macro-based generation isn't exactly super configurable, and this implementation does have the unfortunately side effect that `bitflags_values!` can be used at most once per struct as it adds a new `impl` block each time.

brson:
@Centril indeed a working out-of-tree user would help make the case.

Does extbitflags depend on being compatible with `bitflags!`-generated types? If not it would be an alternative to either duplicate the content of `bitflags!` in extbitflags, or merge extbitflags into bitflags.

Centril:
@alexcrichton 

That could be remedied by splitting `bitflags_values!` into 2 macros, 1 for creating 1 "value", and one for adding the impl block, like: `bitflags_value!` (not, singular) and `bitflags_build!`.

The main problem was that I couldn't do `bitflags!` w/o specifying any values...

@brson 

I'm starting to think that it isnt possible to make the macro work at all since neither:

```rust
//================================================================================
// macro_seqc!
//================================================================================

/**
 * General helper for macros, "returns" the number of arguments passed.
 *
 * Example:
 * ````rust
 * assert_eq!( 3, macro_seqc!( 9, 9, 9 ) );
 * ```
 */
#[macro_export]
macro_rules! macro_seqc {
    ( $( $thing:expr ),+ ) => { macro_seqc_expr!(1, $($thing),*) };
    ( $( $thing:pat ),+ ) => { macro_seqc_pat!(1, $($thing),*) };
    () => { 0 };
}

/**
 * Helper for macro_seqc!
 */
#[macro_export]
macro_rules! macro_seqc_expr {
    ( $id:expr, $head:expr, $( $tail:expr ),* ) => ({
    	macro_seqc_expr!( $id + 1, $( $tail ),* )
    });
    ( $id:expr, $last:expr ) => ( $id );
}

/**
 * Helper for macro_seqc!
 */
#[macro_export]
macro_rules! macro_seqc_pat {
    ( $id:expr, $head:pat, $( $tail:pat ),* ) => ({
    	macro_seqc_pat!( $id + 1, $( $tail ),* )
    });
    ( $id:expr, $last:pat ) => ( $id );
}
#[macro_export]
macro_rules! ebf {
	( $( #[$attr:meta] )* $BitFlags:ident: $T:ty {
		$( $Flag:ident ),+
	} ) => {
		bitflags! {
			$( #[$attr] )*
			flags $BitFlags: $T
		}
		/*
		bitflags_values! { $BitFlags {
			$( const $Flag = 1 << (macro_seqc!( $( $Flag ),+ )) ),+
		} }
		*/
		bitflags_values! { $BitFlags {
			ebf_flags2!( 0, $( $Flag:ident ), + )
		} }
	};
}
```
nor the version out-commented works.
The out-commented doesn't work because the macro system doesn't allow you to do repetition in repetition on the same $Flag, the second doesn't work because you can't use macros in macros :/

If I could get it working, I'd love to merge the idea into this repo in some way, with or without this PR.

dtolnay:
Closing due to inactivity and because I don't think we have gotten anything working that would be able to leverage this.

PR #4, opened 2015-02-09
Fixes the build by explicitly including libcore

shadowmint:
Fixes so it compiles again; not sure why this wasn't broken before, but I presume its to do with feature flag updates.


alexcrichton:
Thanks!

PR #7, opened 2015-05-31
Make a nice Debug impl for each bitflags.

cmr:
This adds a Debug impl where previously there was none, so this is a

[breaking-change]

Trying to actually debug anything using bitflags is otherwise absolutely infuriating.


alexcrichton:
Thanks!

Issue #8, opened 2015-06-01
New Debug impl fails to compile when the bitflags are conditional


CraZySacX:
The following code cause a compilation error when compiled not compiled with feature "C".

```rust
bitflags! {
    flags Flags: u32 {
        const FLAG_A = 0b00000001,
        const FLAG_B  = 0b00000010,
        #[cfg(feature = "C")] const FLAG_C = 0b00000100,
    }
}
```

> <bitflags macros>:12:22: 12:28 error: unresolved name `FLAG_C`
> <bitflags macros>:12 if self . contains ( $ Flag ) {
>                                           ^~~~~~
> <bitflags macros>:1:1: 79:67 note: in expansion of bitflags!
> <bitflags macros>:77:1: 79:63 note: expansion site
> <bitflags macros>:1:1: 79:67 note: in expansion of bitflags!

alexcrichton:
Fixed in #9

PR #9, opened 2015-06-01
re-implemented Debug to show the bits field of the internal struct as…

CraZySacX:
… binary (8-bits wide)

Fixes issue #8 by re-implementing Debug to a simpler representation:

```$BitFlags { bits: 0bXXXXXXXX }```

If this isn't preferable, I would recommend dropping the general Debug impl and let everyone implement their own.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,7 +77,8 @@
 ///     flags.clear();
 ///     assert!(flags.is_empty());
 ///     assert_eq!(format!("{}", flags), "hi!");
-///     assert_eq!(format!("{:?}", FLAG_A | FLAG_B), "FLAG_A | FLAG_B");
+///     assert_eq!(format!("{:?}", FLAG_A | FLAG_B), "Flags { bits: 0b00000011 }");
+///     assert_eq!(format!("{:?}", FLAG_B), "Flags { bits: 0b00000010 }");
 /// }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by alexcrichton** [comment]
Thanks! I think I'll remove the 10-character width limit for now (as it could be larger or smaller), but I'll do that locally.

**Comment by cmr** [comment]
This sucks a lot for my usecases at least, since a non-symbolic representation is impossible to decipher (I have 61 cases right now). I don't see a good way to do it in the macro though in the face of #8.

File: src/lib.rs [hunk 1]
```diff
@@ -139,19 +138,10 @@ macro_rules! bitflags {
...
-                        try!(f.write_str(stringify!($Flag)));
-                    }
-                )+
-                let _ = first; // silence! the last assignment from the above
-                               // expansion doesn't get read. fix that.
-                Ok(())
+                let out = format!("{} {{ bits: {:#010b} }}",
+                                  stringify!($BitFlags),
+                                  self.bits);
+                f.write_str(&out[..])
```
chris-morgan: General design note: there’s no need to create an intermediate string; you should use `write!` in here:

```rust
                write!(f, "{} {{ bits: {:#010b} }}",
                       stringify!($BitFlags),
                       self.bits)
```

PR #17, opened 2015-07-06
Add FromStr implementation

cmr:


alexcrichton:
I [raised some concerns](https://botbot.me/mozilla/rust-libs/2015-07-06/?msg=43828357&page=2) on IRC about how this may be a little too surprising to provide by default, so I'm not sure I'm 100% in favor of this.

dtolnay:
**\<acrichto>** Elaine: hm that seems less "obviously desirable" than much of the other functionality that has been recently added

> **\<Elaine>** I agree.

**\<acrichto>** specifically the Rust syntax you're using to declare the flag names may be fairly orthogonal to the actual names of the flags
**\<acrichto>** and you can always provide the FromStr impl yourself

> **\<Elaine>** "Providing an impl yourself" just doesn't scale when you have dozens of variants.
> **\<Elaine>** I'm not sure what "actual names of the flags" means -- afaict this is the strict inverse of the Display impl?

**\<acrichto>** right but it scales fine if you write your own macro
**\<acrichto>** and I hope there isn't a Display impl
**\<acrichto>** that should be Debug

> **\<Elaine>** Right it's Debug, sorry.

**\<acrichto>** ah yeah
**\<acrichto>** so it's not clear that the "string name" for a flag is the same as the rust identifier name

> **\<Elaine>** I see, yes.
> **\<Elaine>** The nice thing about having bitflags! provide it is that as a user of a library using bitflags!, I get a base feature level instead of having to ask the author to make changes/implementing it myself.
> **\<Elaine>** But I agree that this may be suboptimal.

dtolnay:
I agree with not providing this impl. The "A | B | C" representation is useful for Debug output but is never what I would expect to use with FromStr.

brson:
Closing.

PR #24, opened 2015-11-28
Convert to associated constants

tamird:
Two problems with this:
- breaking API change
- requires nightly

I suppose this will need to wait until associated constants are stable.


alexcrichton:
Yeah for now this crate builds on stable, so we'll have to wait for associated constants to be stable to merge this unfortunately.

phil-opp:
@tamird It seems like the `associated_const` feature will remain unstable for the near future. So it might make sense to pre-release this on crates.io as a separate crate (e.g. `bitflags_associated_consts`). This would allow nightly users to take advantage of the nicer API.

tamird:
This needs a rebase to pick up various changes made on master; I'll try to get to it this week.

tamird:
Rebased; probably won't pass until 1.20 is stable.

tamird:
Rebased again. Beyond the use of associated constants, this also uses compiletest, which requires nightly. Do we want to just remove the compile-fail test?

alexcrichton:
The stable/beta builds seem to still be failing due to experimental features?

tamird:
Yes, I believe that's because this PR adds a dependency on compiletest, which requires nightly (see my last comment). Should I just remove the compile-fail tests?

tamird:
Removed compiletest; lmk if you'd like me to squash the commits, but this should be green now.

alexcrichton:
👍 

Thanks @tamird!

PR #26, opened 2015-12-23
Added clippy support to bitflags, and cleaned up some lints

CraZySacX:
Ran across some lints in another project that depends on bitflags.  Couldn't get them allowed easily, so added clippy as an optional dependency here and cleaned up the lints.


dimbleby:
@CraZySacX - is this pull request in the state that you intended?  It looks as though it contains a bunch of changes not meant for this repository eg renaming the crate...

CraZySacX:
It is not, forgot to close this.

PR #29, opened 2016-01-05
Add support for no_std

### LINKED ISSUE #27: Make no_std once 1.6.0 is out
bitflags really wants to be used in OS-kernel-related code, but since it isn't no_std it can't be.

Amanieu:
This PR adds support for no_std by adding a "no_std" cargo feature. The use of the cargo feature is needed for now because stable rust does not support no_std yet. Once 1.6.0 is released then the feature can be removed and bitflags can be made unconditionally no_std.

Fixes #27

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,20 @@
 
 //! A typesafe bitmask flag generator.
 
+// Compile the crate with no_std if possible. The "no_std" feature can be
+// removed once no_std becomes available in 1.6.0 stable. In the meantime no_std
+// must be disabled by default to allow the crate to work on older rust versions.
+#![cfg_attr(all(feature = "no_std", not(test)), no_std)]
+
+#[cfg(all(feature = "no_std", not(test)))]
+#[macro_use]
+extern crate core as std;
+
+// Re-export libstd/libcore using an alias so that the macros can work in no_std
+// crates while remaining compatible with normal crates.
+#[doc(hidden)]
+pub use std as __core;
+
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -10,6 +10,24 @@
...
 //! A typesafe bitmask flag generator.
 
+// Compile the crate with no_std if possible. The "no_std" feature can be
+// removed once no_std becomes available in 1.6.0 stable. In the meantime no_std
+// must be disabled by default to allow the crate to work on older rust versions.
+#![cfg_attr(feature = "no_std", no_std)]
+
+#[cfg(all(feature = "no_std", test))]
+#[macro_use]
+extern crate std;
```
alexcrichton: You may actually be able to get away with `extern crate std as core` here and avoid the `#[cfg]` below

alexcrichton:
Looks good to me, thanks!

src/lib.rs
Amanieu: Something like this?
```rust
// Compile the crate with no_std if possible. The "no_std" feature can be
// removed once no_std becomes available in 1.6.0 stable. In the meantime no_std
// must be disabled by default to allow the crate to work on older rust versions.
#![cfg_attr(feature = "no_std", no_std)]

#[cfg(all(feature = "no_std", test))]
#[macro_use]
extern crate std;

#[cfg(not(feature = "no_std"))]
extern crate std as core;

// Re-export libcore using an alias so that the macros can work in no_std crates
// while remaining compatible with normal crates.
#[doc(hidden)]
pub use core as __core;
```

PR #30, opened 2016-01-09
Keep clippy happy

dimbleby:
Macro expansion happens before clippy does its thing - probably there's no other way - so code that uses `bitflags` can't be clippy-clean unless `bitflags` is clippy-clean.

When the tweaked variable is called `_first`, clippy complains that it begins with an underscore even though it is actually used.

But on re-naming as `first` we find that the second assignment of the variable isn't always used (presumably in cases where there's only one flag present).  So need to `#[allow(unused_assignments)]` too.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -152,6 +152,7 @@ macro_rules! bitflags {
                 // defined variants, leaving only the undefined ones with the
                 // bit value of 0.
                 #[allow(dead_code)]
+                #[allow(unused_assignments)]
                 mod dummy {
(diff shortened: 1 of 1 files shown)


dimbleby:
Just noticed the overlap with #26 - this is essentially a subset of that.

alexcrichton:
Thanks! For now I'll close this in favor of #26 and we can discuss over there.

dimbleby:
Not sure that's wise - #26 looks kinda crazy, and a long way from being mergeable.  Eg it renames the crate...

PR #32, opened 2016-01-10

fiveop:
The patch implements assignment operators for bitflag structs.

If assignment operators become stable (https://github.com/rust-lang/rust/issues/28235), it can be merged.

Currently, anyone using the `bitflags` macro would need to enable `op_assign_traits` for his crate

```
#![feature(op_assign_traits)]
```

and anyone using the assignment operators would need to enable `augmented_assignments` for his crate

```
#![feature(augmented_assignments)]
```

This includes the tests and doc tests, thus these fail as long as the feature is not stable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,6 @@ documentation = "https://doc.rust-lang.org/bitflags"
 description = """
 A macro to generate structures which behave like bitflags.
 """
+
+[features]
+assignment_operators = []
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! This is a nightly feature, however, and this is a stable crate, so we can't have this turned on by default. I would be ok putting it behind a feature, however.

fiveop:
Open a new Pull Request rebased on current master: #35.

PR #35, opened 2016-01-13
 Implement assignment operators

fiveop:
See Pull Request #32.

PR #36, opened 2016-01-14

fiveop:
See #35 and #32.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - |
-    [ "$TRAVIS_RUST_VERSION" = "1.0" ] || cargo test --verbose --features no_std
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -7,7 +7,11 @@ sudo: false
 script:
   - cargo build --verbose
   - cargo test --verbose
-  - [ "$TRAVIS_RUST_VERSION" = "1.0" ] || cargo test --verbose --features no_std
+  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
+       cargo test --verbose --features no_std
```
alexcrichton: Can you be sure to test `no_std` on non-1.0 as well? It should be tested on beta at least and likely on stable as well.

fiveop: Is there a reason, why we test against 1.0 and not against stable? When that's answered I'll amend the PR.

alexcrichton: Yeah for now we try to get the most coverage possible (e.g. keep this back to 1.0 while possible)

PR #37, opened 2016-01-15

fiveop:
See #32 , #35 and #36.


fiveop:
I was just told, that no_std will be in stable next week, so we can add stable as a fourth build version then.

alexcrichton:
Thanks!

Also feel free to just push to branches in the future, no need for new PRs :)

PR #38, opened 2016-01-16
Add private bitflags

### LINKED ISSUE #25: Bitflags should be private by default
Bitflags should match the semantics of structs, which are private by default unless explicitly exported using `pub`.

Since this is a compat-breaking change, it would make sense to do it at the same time as the switch to associated constants and namespaced flags (#24).

Example:
```rust
bitflags! {
    pub flags Flags: u32 {
        const FLAG_A   = 0b00000001,
        const FLAG_B   = 0b00000010,
    }
}
```

Amanieu:
This adds the ability to define private bitflags that aren't exported outside of the current module using `priv flags`.

Ideally this would be done by matching the existing behavior in the language: making private bitflags the default behavior and requiring the use of `pub flags` to export it outside the mode. However there are a [large number](https://crates.io/crates/bitflags/reverse_dependencies) of crates that use `bitflags` with a `*` version, which makes me somewhat nervous to break them all.

Fixes #25

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,6 +101,35 @@ pub use std as __core;
 /// }
 /// ```
 ///
+/// # Visibility
+///
+/// The generated struct and its associated flag constants are not exported
+/// out of the current module by default. A definition can be exported out of
+/// the current module by adding `pub` before `flags`:
+///
+/// ```{.rust},ignore
+/// #[macro_use]
+/// extern crate bitflags;
+///
+/// mod example {
+///     bitflags! {
+///         pub flags Flags1: u32 {
+///             const FLAG_A   = 0b00000001,
+///         }
+///     }
+///     bitflags! {
+///         flags Flags2: u32 {
+///             const FLAG_B   = 0b00000010,
+///         }
+///     }
+/// }
+///
+/// fn main() {
+///     let flag1 = example::FLAG_A;
+///     let flag2 = example::FLAG_B; // error: const `FLAG_B` is private
+/// }
+/// ```
+///
 /// # Attributes
 ///
 /// Attributes can be attached to the generated `struct` by placing them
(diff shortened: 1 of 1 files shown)


alexcrichton:
I agree that private is the better default (e.g. matches Rust) and requiring an explicit `pub` is the best way to go here, so I'd personally prefer to go in that direction.

Amanieu:
I changed it to private by default and requiring an explicit `pub`. However I am still worried that this will break a lot of crates that use `bitflags` with a `*` version: https://crates.io/crates/bitflags/reverse_dependencies

alexcrichton:
Ah yeah it's true that this may break some crates, but `*` dependencies are no longer allowed on crates.io and those are ones that probably need to be updated anyway. I wouldn't necessarily want to be locked in to prevent *any* breaking changes so it seems fine to do this.

Thanks again @Amanieu!

gkoz:
There's a confusing situation now: 0.4.0 from git master and the documentation at `doc.rlo` have default private but the latest crates.io 0.4.0 package has default public.

Amanieu:
This is because this PR hasn't been pushed to crates.io yet.

alexcrichton:
I've published 0.5.0 now with this change.

xitep:
hello, it looks like after this change bitflags are _forced_ to standard types, e.g. u32. is this an accident or purpose?

the `lmdb-rs` project - which i tried to upgrade - uses amongst others [libc::c_uint in its bitflags definitions.](https://github.com/vhbit/lmdb-rs/blob/master/src/core.rs#L183) but i see no possibility to include the corresponding `use libc::c_uint` declaration in the newly generated `dummy` module.

any advice for me?

Amanieu:
@xitep I think the best advice would be to just use `u32` for now. `libc::c_uint` is going to be equivalent to that type on most sane platforms anyways.

@alexcrichton I tried fixing this but couldn't figure out a way to reliably pass the type down to the dummy module (`type` doesn't seem to be covered by macro hygiene). Note that the associated constants version (#24) doesn't have this issue since it doesn't use dummy modules.

alexcrichton:
Can't you use an absolute path like `::libc::c_uint`?

xitep:
@alexcrichton thank you! yes, absolute reference make it!

is this the long term plan or do we plan for another solution?

alexcrichton:
Probably, this library is in a relatively bad state unfortunately. Can you open an issue on this?

PR #45, opened 2016-04-09
no_std is stable now

Ericson2314:
Not sure if we still want `__core`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,6 @@ sudo: false
 script:
   - cargo build --verbose
   - cargo test --verbose
-  - cargo test --verbose --features no_std
-  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
-       cargo test --verbose --features "no_std";
-    fi
   - cargo doc
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,3 @@ documentation = "https://doc.rust-lang.org/bitflags"
 description = """
 A macro to generate structures which behave like bitflags.
 """
-
-[features]
-no_std = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,20 +10,17 @@
 
 //! A typesafe bitmask flag generator.
 
-// Compile the crate with no_std if possible. The "no_std" feature can be
-// removed once no_std becomes available in 1.6.0 stable. In the meantime no_std
-// must be disabled by default to allow the crate to work on older rust versions.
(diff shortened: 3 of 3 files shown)


Amanieu:
We still need `__core` because we don't know whether the crate using `bitflags!` has `core` or `std`, and it could even have overridden these names to mean something else entirely. Therefore the only reliable way of referring to `core` is through `$crate::__core`.

Amanieu:
However note that removing the `no_std` Cargo feature is a breaking change. We could leave it as a no-op or remove it and bump our version number.

Ericson2314:
Hmm I would just bump it. Don't think this crate is ready for 1.0 and less churn until associated types are stable anyways.

Also, is `pub extern crate` a thing? Can we do that with `$crate::core` and drop the underscores?

Amanieu:
I think `pub extern crate` currently gives a warning saying it doesn't work properly and that you should use `pub use` instead.

The reason I added underscores was because `__core` is not part of the API of this crate, it only needs to be publicly exported so that it can be used by the macro.

Ericson2314:
Alright, makes sense. I'll try `pub extern crate core as __core`, and bump the version.

Ericson2314:
https://github.com/rust-lang/rust/issues/26775 Ok `pub extern crate` is currently broken.

Ericson2314:
Nevermind, it does work after all. Used that, and bumped version number, so this should be ready to go.

Amanieu:
Travis doesn't seem to agree (see below)

Ericson2314:
Ah, I only tested on nightly oops. I'll revert that, and fix the Travis error on the other two where it tries to build with the no_std feature.

phil-opp:
I think you can just do a:

```
pub use core as __core;
```

The `extern crate core` is not needed since it's imported automatically when using `#![no_std]`.

Ericson2314:
@phil-opp It's needed for tests when `#![no_std]` is disabled.

alexcrichton:
Looks like CI is failing? Otherwise seems like a good change to me!

Ericson2314:
Ok, fixed problems. Also made it always use `#![no_std]` and conditionally import std for tests, which seems more idiomatic.

alexcrichton:
Can you back out the version bump? That's typically done as a separate commit

Ericson2314:
Separate PR I assume you mean? Unless I messed up it is a separate commit ATM.

alexcrichton:
Yes, separate PR (or I'll just do it when I merge this)

Ericson2314:
Sure. I'll just remove the commit then.

Ericson2314:
Done!

PR #47, opened 2016-04-20
Remove the assignment_operators feature, its requirements are now stable

MJDSys:
With rust 1.8, augmented_assignments and op_assign_traits are stable.  Thus
just enable them globally.


alexcrichton:
Nice!

PR #49, opened 2016-04-23
Implement Extend for bitflags

kamalmarhubi:
This commit adds an `Extend` impl, and redefines the `FromIterator` impl
to use it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,8 +135,9 @@ pub use core as __core;
 /// Additional traits can be derived by providing an explicit `derive`
 /// attribute on `flags`.
 ///
-/// The `FromIterator` trait is implemented for the `struct`, too, calculating
-/// the union of the instances of the `struct` iterated over.
+/// The `Extend` and `FromIterator` traits are implemented for the `struct`,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -452,12 +453,19 @@ macro_rules! bitflags {
...
+        impl $crate::__core::iter::Extend<$BitFlags> for $BitFlags {
+            fn extend<T: $crate::__core::iter::IntoIterator<Item=$BitFlags>>(&mut self, iterator: T) {
+                for item in iterator {
+                    self.insert(item)
+                }
+            }
+        }
+
         impl $crate::__core::iter::FromIterator<$BitFlags> for $BitFlags {
+            #[inline]
```
alexcrichton: It should be fine to remove this `#[inline]` because the function is already generic which will inline it elsewhere anyway

alexcrichton:
Neat!

src/lib.rs
kamalmarhubi: Ah good to know, thanks! (Done.)

PR #50, opened 2016-04-27
Recommend current bitflags crate version in README.md

alexander255:
Yes, that's the only change! Build errors (missing `pub`) occurred with the currently recommended `0.3` version.


alexcrichton:
Thanks!

PR #55, opened 2016-12-22
Add .set method (condition-based insert/remove)

mehcode:
```rust
let f = FlagA | FlagC;
f.set(FlagB, true);  // equivalent to f.insert(FlagB)
f.set(FlagA, false);  // equivalent to f.remove(FlagA)
```

This adds a convenience method to explicitly set the value of a bit. I, at least, use this pattern heavily in https://github.com/mehcode/wadatsumi (A gameboy emulator) so I'm proposing this.


alexcrichton:
Thanks!

Issue #56, opened 2017-01-28
Support for nightly 128-bit integers


astocko:
What do you think about adding feature gated 128-bit based bitflags? While trying u128 in nightly, I noticed that bitflags silently degrades to u64.

I can work on a PR if interested.

alexcrichton:
This seems fine to support so long as it doesn't cause the crate to require nightly

astocko:
Added support in #57.

PR #57, opened 2017-02-11
Add u128 feature for complete u128 compatibility

astocko:
Added a u128 feature with a feature gated bitflags width type for use in the macro. Enables functionality like debug formatting of flags that exceed 64 bits. #56


alexcrichton:
This seems to only be used in the formatting pieces, and there could we just use `$T` instead of an explicit size? That way we could defer this decision entirely to downstream crates I think?

astocko:
Implemented as suggested. The change does require absolute paths for custom types and type aliases.

alexcrichton:
Oh right sorry I forgot about that :(, that's actually a pretty crucial reason for this feature.

In that case I guess we'll have to stick with an unstable feature

astocko:
I will revert to my original PR. Any comments or suggestions?

alexcrichton:
Looks like this may still be the requires-absolute-paths version?

alexcrichton:
Thanks!

dtolnay:
#61 reimplements this in a different way that neither requires absolute paths nor a cfg to opt into u128 compatibility.

PR #74, opened 2017-03-08
Function-local bitflags (another approach)

### LINKED ISSUE #71: Function-local bitflags
Is there any hope of supporting usage like this? The error here is inconsistent with being able to declare structs and other types of items local to a function.

```rust
#[macro_use]
extern crate bitflags;

fn main() {
    bitflags! {
        flags Flags: u8 {
            const A = 1,
            const B = 2,
        }
    }
    println!("{:?}", A);
}
```

```
error[E0425]: cannot find value `A` in module `super::super`
  --> src/main.rs:5:5
   |
5  |       bitflags! {
   |  _____^ starting here...
6  | |         flags Flags: u8 {
7  | |             const A = 1,
8  | |             const B = 2,
9  | |         }
10 | |     }
   | |_____^ ...ending here: not found in `super::super`
   |
   = note: this error originates in a macro outside of the current crate
```

dtolnay:
Fixes #71 in a different way than #73.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,63 +216,46 @@ macro_rules! bitflags {
         impl $crate::__core::fmt::Debug for $BitFlags {
             fn fmt(&self, f: &mut $crate::__core::fmt::Formatter) -> $crate::__core::fmt::Result {
                 // This convoluted approach is to handle #[cfg]-based flag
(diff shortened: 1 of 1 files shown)


alexcrichton:
This seems slightly less sneaky than macros defining macros, so I think I prefer this slightly. Thanks so much!

PR #83, opened 2017-03-16
s/$BitFlags/Self/g

tamird:
@alexcrichton

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -243,7 +243,7 @@ macro_rules! bitflags {
 
                 let mut first = true;
                 $(
-                    if <$BitFlags as __BitFlags>::$Flag(self) {
+                    if <Self as __BitFlags>::$Flag(self) {
                         if !first {
                             try!(f.write_str(" | "));
                         }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Does this affect the generated documentation?

tamird:
Doesn't seem like it does. The string "$BitFlags" doesn't appear in the documentation outside of the macro definition, which is unchanged here.

tamird:
Well, it does appear in `doc/src/bitflags/lib.rs.html` (where it is now replaced with "Self"), but that's expected since that's the source itself.

alexcrichton:
Oh sorry I meant moreso the generated docs for the bitflags struct that this generates. I'd expect the identifier of the struct to be much more readable than `Self` (e.g. more obvious at a glance), so if there's not a technical reason for this I'd err on the side of more self-explanatory docs.

If there's a technical reason to do this, though, please let me know!

tamird:
Well, it does replace the struct name with "Self". However, isn't that true of all implementations that use Self? If documentation is considered clearer with the explicit type name, perhaps that's a transformation that should be applied to all docs?

alexcrichton:
I would personally consider `Self` less readable than the struct name itself. Sure I can always remember to perform the transformation, but it seems better if I don't have to do it in the first place.

Can you elaborate on the motivation for this PR?

tamird:
Sorry, I didn't mean to suggest a mental transformation - I meant that it should be possible to universally apply the transformation at the docs generation stage.

The motivation is just to reduce duplication.

alexcrichton:
Yeah definitely rustdoc *could* do that, but it doesn't currently afaik :(

In that sense would you be ok closing until that's a feature in rustdoc?

tamird:
Sure thing.

PR #85, opened 2017-03-22
Modify Debug implementation to use "(empty)" instead of "" for empty flags

### LINKED ISSUE #64: Empty bitflags has unhelpful Debug representation
The Debug representation is the empty string. This is not very debuggable. I think "(empty)" would be more helpful.

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    pub flags Flags: u8 {
        const A = 0,
    }
}

fn main() {
    println!("{:?}", Flags::empty());
}
```

crazymerlyn:
Fixes #64.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,8 +252,7 @@ macro_rules! bitflags {
                     }
                 )+
                 if first {
-                    // TODO: should not be the empty string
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #87, opened 2017-03-22
Use semicolons instead of commas.

### LINKED ISSUE #65: Semicolons
It may be too late for such a change to make sense but typically "const" items are followed by a semicolon, not comma.

Semicolon would be more consistent with lazy_static as well.

```rust
bitflags! {
    flags Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
    }
}

const C: u32 = 0b00000100;
const D: u32 = 0b00001000;

lazy_static! {
    static ref E: u32 = 0b00010000;
    static ref F: u32 = 0b00100000;
}
```

topecongiro:
Closes #65.


dtolnay:
This is a bit more complicated because commas are treated as separators with an optional trailing separator, while semicolons need to be treated as terminators.

```rust
{ a, b, c } // ok
{ a, b, c, } // ok
{ a; b; c; } // ok
{ a; b; c } // not ok
```

The pairs of rules that look like this:

```rust
$($(#[$Flag_attr:meta])* const $Flag:ident = $value:expr);+
$($(#[$Flag_attr:meta])* const $Flag:ident = $value:expr);+;
```

need to be replaced by a single rule that looks like this:

```rust
$($(#[$Flag_attr:meta])* const $Flag:ident = $value:expr ;)+
```

topecongiro:
@dtolnay Thank you for your comment. I will change my PR accordingly.

alexcrichton:
Looks great to me, thanks @topecongiro! Want to rebase and I'll merge?

topecongiro:
@alexcrichton Thank you! Rebased.

Issue #66, opened 2017-03-07
Mention Default trait in the docs


dtolnay:
It makes sense to me why Default is not derived for bitflags - you may want a default value that is different from the empty value. Since Default is such a common trait, it would be helpful to mention explicitly in the "Trait implementations" section of the rustdoc and show an example using `#[derive(Default)]`.

PR #89, opened 2017-03-23
Example for Default trait in the docs

jptomanelli:
This is for #66 
Please let me know if this looks good!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,13 +84,21 @@ pub use core as __core;
 ///     }
 /// }
 ///
+/// impl Default for Flags {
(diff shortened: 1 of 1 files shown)


dtolnay:
I think people can figure out how to handwrite an implementation of Default. #66 is asking to show how to derive Default for bitflags. Also please add a sentence or two of explanation after "The generated \`struct\`s can also be extended with type and trait implementations."

```rust
bitflags! {
    #[derive(Default)]
    flags Flags: u32 {
        const FLAG_A = 0b00000001,
        const FLAG_B = 0b00000010,
    }
}
```

Issue #78, opened 2017-03-16
Add html_root_url crate attribute


brson:
Also change the Cargo.toml 'documentation' key to "https://docs.rs/bitflags".

brson:
For now follow the example [here](https://github.com/BurntSushi/byteorder/pull/78) where I used scare-comments to indicate that the crate version and the html_root_url need to stay in sync.

brson:
Thanks @Cldfire !

PR #94, opened 2017-03-29
Update documentation links and attributes

Cldfire:
Fulfills #78 

I also updated the `Documentation` link in the README as I'm fairly sure it should point to the same documentation as the Cargo.toml.

Now, in doing this, I noticed that the current version in the Cargo.toml of this repo (0.8.0) **does not match** the version of bitflags that is currently published as a crate (0.8.2). I made the `html_root_url` attribute link to version 0.8.2 on docs.rs.

Is this version mismatch intended? If not I can add another commit to this PR to fix it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,17 @@
 [package]
 
 name = "bitflags"
-version = "0.8.0" # also update number in readme for breaking changes
+# NB: When modifying, also modify:
+#   1. html_root_url in lib.rs
+#   2. number in readme (for breaking changes)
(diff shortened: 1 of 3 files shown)


alexcrichton:
Ah yeah there are tags for 0.8.1 and 0.8.2 for various reasons, so the mismatch is intended. Thanks for the pr though! Want to rebase and I'll merge?

Cldfire:
Rebased.

alexcrichton:
Thanks!

Issue #75, opened 2017-03-16
Move docs to the crate level


brson:
This crate presently has no crate level docs, and that is the most important place for docs to be.

Let's move the current content of the `bitflags!` rustdocs up to the crate level, leaving behind copies of the examples, and a hyperlink to the crate level.

PR #96, opened 2017-03-29
Move macro docs to the crate level

Cldfire:
Fulfills both #75 and #79

Is there a better way to link to the crate-level docs? I'm not too familiar with linking within rustdoc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,9 +139,6 @@
 //! - `Sub` and `SubAssign`: set difference
 //! - `Not`: set complement
 //!
-//! As long as the assignment operators are unstable rust feature they are only
-//! available with the crate feature `assignment_ops` enabled.
-//!
 //! # Methods
 //!
 //! The following methods are defined for the generated `struct`:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,159 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! A typesafe bitmask flag generator.
+//! A typesafe bitmask flag generator useful for sets of C-style bitmask flags.
+//! It can be used for creating typesafe wrappers around C APIs.
+//!
+//! The `bitflags!` macro generates a `struct` that manages a set of flags. The 
+//! flags should only be defined for integer types, otherwise unexpected type 
+//! errors may occur at compile time.
+//!
+//! # Example
+//!
+//! ```{.rust}
+//! #[macro_use]
+//! extern crate bitflags;
+//!
+//! bitflags! {
+//!     struct Flags: u32 {
+//!         const FLAG_A       = 0b00000001;
+//!         const FLAG_B       = 0b00000010;
+//!         const FLAG_C       = 0b00000100;
+//!         const FLAG_ABC     = FLAG_A.bits
+//!                            | FLAG_B.bits
+//!                            | FLAG_C.bits;
+//!     }
+//! }
+//!
+//! fn main() {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -8,7 +8,156 @@
...
+//! A typesafe bitmask flag generator useful for sets of C-style bitmask flags.
+//! It can be used for creating typesafe wrappers around C APIs.
+//!
+//! The `bitflags!` macro generates a `struct` that manages a set of flags. The 
+//! flags should only be defined for integer types, otherwise unexpected type 
+//! errors may occur at compile time.
+//!
+//! # Example
+//!
+//! ```{.rust}
```
steveklabnik: don't include the `{.rust}` bit; this is an older style, today it would be

    ```rust

but, we don't include the `rust` part in API docs, so just triple graves :smile:

src/lib.rs
```diff
@@ -8,7 +8,156 @@
...
+//!     assert_eq!((e1 & e2), FLAG_C);     // intersection
+//!     assert_eq!((e1 - e2), FLAG_A);     // set difference
+//!     assert_eq!(!e2, FLAG_A);           // set complement
+//! }
+//! ```
+//!
+//! The generated `struct`s can also be extended with type and trait
+//! implementations:
+//!
+//! ```{.rust}
```
steveklabnik: same here

src/lib.rs
```diff
@@ -8,7 +8,156 @@
...
+//! }
+//! ```
+//!
+//! # Visibility
+//!
+//! The generated struct and its associated flag constants are not exported
+//! out of the current module by default. A definition can be exported out of
+//! the current module by adding `pub` before `flags`:
+//!
+//! ```{.rust},ignore
```
steveklabnik: just 

    ```ignore

src/lib.rs
```diff
@@ -27,11 +176,9 @@ extern crate std;
 #[doc(hidden)]
 pub use core as __core;
 
-/// The `bitflags!` macro generates a `struct` that holds a set of C-style
-/// bitmask flags. It is useful for creating typesafe wrappers for C APIs.
+/// The macro used to generate the flag structure.
 ///
-/// The flags should only be defined for integer types, otherwise unexpected
-/// type errors may occur at compile time.
+/// See the [crate level docs](https://docs.rs/bitflags) for complete documentation.
```
steveklabnik: you had mentioned this aspect of it. The issue here is that this breaks locally; use a relative link instead. I believe that that means just `../index.html`, but I didn't 100% verify this

Cldfire:
@steveklabnik, thank your for the comments. `../index.html` is not working for me locally, but `../bitflags/index.html` is. Should I commit the latter?

src/lib.rs
Cldfire: Whoops, I should comment here ^^

`../index.html` did not work for me locally, but `../bitflags/index.html` did. Is it fine if I add a commit for the latter?

src/lib.rs
Cldfire: Gotcha :smile:

src/lib.rs
steveklabnik: ah yeah, makes sense. 👍

Cldfire:
There, I believe that should take care of those issues.

Cldfire:
Oh, and should I remove the `unstable_testing` feature from `Cargo.toml` as well? It appears it is no longer relevant, going along with the removal of mention of `assignment_ops` being a nightly feature from the docs.

alexcrichton:
Thanks @Cldfire!

Issue #81, opened 2017-03-16
Add an example of what the macro-expanded API looks like


brson:
This crate has a documentation problem in that it generates a bunch of methods, and they are hard to document. I believe the way we want to address this is by creating another module, containing a hand-coded API that looks like the bitflags produced API, and treat that entire module as documentation, sort of like error-chain does.

Will take some experimentation.

Cldfire:
I will work on this later today if it has not been completed by then.

koivunej:
Noticed that this was free and opened #99.

PR #97, opened 2017-03-29
Proposal for example of generated macro code

Cldfire:
This is aimed at fulfilling #81.

I am making this PR as a sort of initial proposal (hence the title) and am looking for feedback. I used `cargo-expand` to copy + paste the result of macro expansion into a file and then manually prettified it to turn it into what it is now, which I believe is what the issue description desired.

Some questions I can think of that will need to be answered:

* What additional documentation (on top of what is already provided from the macro expansion) should be written for this example?
* What other tests should there be to ensure that the example does not differ from the macro?
* Are the few tests that I currently have thrown together meaningful?

I also ran into a bit of an issue trying to use the `bitflags!` macro from the new file (`example_generated.rs`), which is why the example tests are not in that file and the doc test for the example struct is currently ignored. If someone could help me out with that it would be much appreciated :smile:


src/example_generated.rs
```diff
@@ -0,0 +1,330 @@
+//! This module contains an example of code generated by the macro. **IT IS NOT MEANT
+//! TO BE USED OUTSIDE THIS CRATE**.
+//!
+//! If you would like to see what the actual code looks like, click the `src` button
+//! located top-right. If you are not interested in the code and simply want to
+//! browse the functionality exposed to you, click on the `ExampleFlags` struct and
+//! read on.
+
+use __core::cmp::Ordering;
+use __core::hash::{Hash, Hasher};
```
Cldfire: Use statements on line 9 and 10 here are not used. Marking so that I remember to remove them.

src/example_generated.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Returns the set difference of the two sets of flags.
+    #[inline]
+    fn sub(self, other: ExampleFlags) -> ExampleFlags {
+        ExampleFlags { bits: self.bits & !other.bits }
+    }
+}
+
+impl SubAssign for ExampleFlags {
+    /// Disabled all flags enabled in the set.
```
Cldfire: Disabled -> Disables

alexcrichton:
Thanks for the PR @Cldfire! I think though the purpose of the issue was to use the macro invocation itself to generate this example rather than expand it by hand. You can see an example in [error-chain](https://github.com/brson/error-chain/blob/master/src/example_generated.rs).

Cldfire:
I was wondering about that. The text "a hand-coded API that looks like the bitflags produced API" from this issue eventually led me to believe that this was what was desired; I guess I was wrong.

In any case, feel free to close this PR.

alexcrichton:
Ok, thanks for the PR regardless @Cldfire! If you'd like to send a PR with the macro-generated example it'd be more than welcome as well!

PR #99, opened 2017-04-04
Example generated

### LINKED ISSUE #81: Add an example of what the macro-expanded API looks like

koivunej:
This is an [`error_chain`](https://github.com/brson/error-chain/blob/master/src/example_generated.rs) style example_generated.rs guarded by an on-by-default feature (`example_generated`) also as in `error_chain`.

The crate level example is used as the example.

Fixes #81.


brson:
Thanks @koivunej this looks great to me. @rust-lang-nursery/libs does this look like the right pattern to you?

alexcrichton:
Looks good to me!

brson:
Thanks @koivunej !

PR #101, opened 2017-04-11
Add examples of deriving and implementing `Default`, fixes #66

### LINKED ISSUE #66: Mention Default trait in the docs

robo9k:
This is more than just mentioning in that it shows both how to derive `Default` as well as implementing it oneself, such as in #89. Turned out to be a bit verbose/prominent due to the long-ish code examples.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -158,6 +158,60 @@
 //! - `remove`: removes the specified flags in-place
 //! - `toggle`: the specified flags will be inserted if not present, and removed
 //!             if they are.
+//!
+//! ## Default
+//!
+//! The `Default` trait is not automatically implemented for the generated struct.
+//!
+//! If your default value is equal to `0` (which is the same value as calling `empty()`
+//! on the generated struct), you can simply derive `Default`:
+//!
+//! ```
+//! #[macro_use]
+//! extern crate bitflags;
+//!
+//! bitflags! {
+//!     // Results in default value with bits: 0
+//!     #[derive(Default)]
+//!     struct Flags: u32 {
+//!         const FLAG_A       = 0b00000001;
+//!         const FLAG_B       = 0b00000010;
+//!         const FLAG_C       = 0b00000100;
+//!     }
+//! }
+//!
+//! fn main() {
+//!     let derived_default: Flags = Default::default();
+//!     assert_eq!(derived_default.bits(), 0);
+//! }
+//! ```
+//!
+//! If your default value is not equal to `0` you need to implement `Default` yourself:
+//!
+//! ```
+//! #[macro_use]
+//! extern crate bitflags;
+//!
+//! bitflags! {
+//!     struct Flags: u32 {
+//!         const FLAG_A       = 0b00000001;
+//!         const FLAG_B       = 0b00000010;
+//!         const FLAG_C       = 0b00000100;
+//!     }
+//! }
+//!
+//! // explicit `Default` implementation
+//! impl Default for Flags {
+//!     fn default() -> Flags {
+//!         FLAG_A | FLAG_C
+//!     }
+//! }
+//!
+//! fn main() {
+//!     let implemented_default: Flags = Default::default();
+//!     assert_eq!(implemented_default, (FLAG_A | FLAG_C));
+//! }
+//! ```
 
 #![no_std]
 


robo9k:
The example given by @dtolnay in #89 is
```rust
bitflags! {
    #[derive(Default)]
    flags Flags: u32 {
        const FLAG_A = 0b00000001,
        const FLAG_B = 0b00000010,
    }
}
```

I'm not sure it makes sense to have a default value of `0` that is not a flag item itself. If the default value is empty (and does not have another "name"), then just tell the user to call `empty()`? That's what I found confusing about the example above.

dtolnay:
I don't think there is any significance to 0 being a flag item. Think of bitflags as representing a set rather than an enum.

```rust
// more like this
hashset![]
hashset![FLAG_A]
hashset![FLAG_B]
hashset![FLAG_A, FLAG_B]

// less like this
enum Robo9k { FLAG_A, FLAG_B }
Robo9k::FLAG_A
Robo9k::FLAB_B
```

Standard library set types have a default which is the empty set.

robo9k:
Okay, got it. Should I change the derive example then?
Currently it's rather focused on showing the difference to the implementation example with my assumption of the default being a flag itself. I would suggest these examples instead:
```rust
bitflags! {
    // Default value with bits: 0
    #[derive(Default)]
    flags Flags: u32 {
        const FLAG_A = 0b00000001,
        const FLAG_B = 0b00000010,
    }
}
```
```rust
bitflags! {
    flags Flags: u32 {
        const FLAG_A = 0b00000001,
        const FLAG_B = 0b00000010,
    }
}

impl Default for Flags {
    fn default() -> Flags {
         // Default value with combination of flags
         Flags { bits: FLAG_A.bits | FLAG_B.bits }
     }
}
```
With this we'd have an empty default as well as a subset of valid flags as another default. Maybe this is distracting and `bits` for the former example should just be some "random" value like `0xDEADBEEF`?

robo9k:
I've changed the examples to those I proposed in my previous comment.
Is there any way to preview how the result will look like on GitHub itself, without copy & pasting the markdown into the comment such as below?

---

## Default

The `Default` trait is not automatically implemented for the generated struct.

If your default value is equal to `0` (which is the same value as calling `empty()`
on the generated struct), you can simply derive `Default`:

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    // Results in default value with bits: 0
    #[derive(Default)]
    struct Flags: u32 {
        const FLAG_A       = 0b00000001;
        const FLAG_B       = 0b00000010;
        const FLAG_C       = 0b00000100;
    }
}

fn main() {
    let derived_default: Flags = Default::default();
    assert_eq!(derived_default.bits(), 0);
}
```

If your default value is not equal to `0` you need to implement `Default` yourself:

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Flags: u32 {
        const FLAG_A       = 0b00000001;
        const FLAG_B       = 0b00000010;
        const FLAG_C       = 0b00000100;
    }
}

// explicit `Default` implementation
impl Default for Flags {
    fn default() -> Flags {
        Flags { bits: FLAG_A.bits | FLAG_C.bits }
    }
}

fn main() {
    let implemented_default: Flags = Default::default();
    assert_eq!(implemented_default, (FLAG_A | FLAG_C));
}
```

src/lib.rs
```diff
@@ -158,6 +158,60 @@
...
+//!         const FLAG_A       = 0b00000001;
+//!         const FLAG_B       = 0b00000010;
+//!         const FLAG_C       = 0b00000100;
+//!     }
+//! }
+//!
+//! // explicit `Default` implementation
+//! impl Default for Flags {
+//!     fn default() -> Flags {
+//!         Flags { bits: FLAG_A.bits | FLAG_C.bits }
```
alexcrichton: I think this can just be `FLAG_A | FLAG_C`, right? (avoids looking at `bits`)

robo9k: Yes. That's mostly a leftover from me wondering about returning random `bits: 0xDEADBEEF` instead of some flag combination. Will update the PR.

brson:
Thanks @robo9k !

PR #105, opened 2017-05-29
Fix the implementation of Formatting traits when other formatting traits were present in scope.

CensoredUsername:
The formatting impls did not explicitly specify which trait was being invoked, which lead to conflicting implementation errors if any other formatting trait was present in scope before the bitflags! invocation (even just a `use std::fmt::Display` beforehand was enough to cause compilation failures.


dtolnay:
I published this in bitflags 0.9.1.

CensoredUsername:
Thanks for the quick response!

PR #107, opened 2017-05-30
Add release notes to README.md

mbrubeck:
These release notes are based on the git commit logs. They may be incomplete. They start at version 0.7.0 because that is the first release whose start and end are marked by git tags.

I wasn't sure which changes in 0.7.0 <s>and 0.8.0</s> were breaking changes, but I'd like to add that information if possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,3 +22,46 @@ and this to your crate root:
 #[macro_use]
 extern crate bitflags;
 ```
+
+## Release notes
+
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks, I put these on the releases page and added a link in 6abb248c15ffff8923a5bd23ec327a77d66dc25b.

PR #114, opened 2017-08-16
Allow empty bitflags

### LINKED ISSUE #113: Support empty bitflags
In `nix` we have some datatypes that we created with `bitflags!`, but since they're cross-platform, not all platforms support all flags. In fact, some platforms support none. Instead of having different declarations of the entire type per-target, it'd be nicer if `bitflags!` supported no entries. For example, then we could implement the following and it would compile on even non-Linux targets:

```rust
bitflags!(
    pub struct SockFlag: c_int {
        #[cfg(target_os = "linux")]
        SOCK_NONBLOCK;
        #[cfg(target_os = "linux")]
        SOCK_CLOEXEC;
    }
);
```

Susurrus:
Closes #113.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -311,7 +311,7 @@ macro_rules! bitflags {
             $(
                 $(#[$inner:ident $($args:tt)*])*
                 const $Flag:ident = $value:expr;
-            )+
+            )*
         }
     ) => {
         #[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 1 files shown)


dtolnay:
The conditionally empty case [already works](https://play.rust-lang.org/?gist=b3c568faf292a8548d75e29e9adbd76f). What is the use case for unconditionally empty bitflags?

Susurrus:
Whoops, looks like we had some outdated comments in our `nix` code and I got ahead of myself for the conditionally-empty case. As for the empty bitflags case, it's come up in `nix` where some APIs we're wrapping define a bitfield that has no elements. We've thought about using an empty bitflag type as the input type so the addition of flags will not be a backwards-incompatible change. We've since changed our minds and left out those fields altogether.

So I think this is unncessary, sorry about the noise.

PR #117, opened 2017-09-09
Add compile-fail tests for private associated constants

tamird:
r? @alexcrichton 

This wanted to be in #24, but I didn't know how to do it.

PR #118, opened 2017-09-09
Avoid re-exporting libcore, remove the need to link

tamird:
r? @alexcrichton this should fix the issue described in https://github.com/rust-lang/rust/pull/44441#issuecomment-328236190.

Note that there is slight merge skew between this and #117; `#[no_link]` should be added to the test being added there, in whichever PR gets merged last.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ and this to your crate root:
 
 ```rust
 #[macro_use]
+#[no_link]
 extern crate bitflags;
 ```
 
--- a/compiletest/tests/compile-fail/private_flags.rs
+++ b/compiletest/tests/compile-fail/private_flags.rs
@@ -1,4 +1,5 @@
 #[macro_use]
+#[no_link]
 extern crate bitflags;
 
 mod example {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,9 @@
 //!
 //! ```
 //! #[macro_use]
+//! #[no_link]
 //! extern crate bitflags;
+//! extern crate core;
 //!
 //! bitflags! {
 //!     struct Flags: u32 {
@@ -50,7 +52,9 @@
 //!
 //! ```
 //! #[macro_use]
+//! #[no_link]
 //! extern crate bitflags;
+//! extern crate core;
 //!
 //! use std::fmt;
 //!
@@ -92,7 +96,9 @@
 //!
 //! ```ignore
 //! #[macro_use]
+//! #[no_link]
 //! extern crate bitflags;
+//! extern crate core;
 //!
 //! mod example {
 //!     bitflags! {
@@ -172,7 +178,9 @@
 //!
 //! ```
 //! #[macro_use]
+//! #[no_link]
 //! extern crate bitflags;
+//! extern crate core;
 //!
 //! bitflags! {
 //!     // Results in default value with bits: 0
@@ -194,7 +202,9 @@
 //!
 //! ```
 //! #[macro_use]
+//! #[no_link]
 //! extern crate bitflags;
+//! extern crate core;
 //!
 //! bitflags! {
 //!     struct Flags: u32 {
@@ -225,11 +235,6 @@
 #[macro_use]
 extern crate std;
 
-// Re-export libcore using an alias so that the macros can work without
-// requiring `extern crate core` downstream.
-#[doc(hidden)]
-pub extern crate core as _core;
-
 /// The macro used to generate the flag structure.
 ///
 /// See the [crate level docs](../bitflags/index.html) for complete documentation.
@@ -238,7 +243,9 @@ pub extern crate core as _core;
 ///
 /// ```
 /// #[macro_use]
+/// #[no_link]
 /// extern crate bitflags;
+/// extern crate core;
 ///
 /// bitflags! {
 ///     struct Flags: u32 {
@@ -266,7 +273,9 @@ pub extern crate core as _core;
 ///
 /// ```
 /// #[macro_use]
+/// #[no_link]
 /// extern crate bitflags;
+/// extern crate core;
 ///
 /// use std::fmt;
 ///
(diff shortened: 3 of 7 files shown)


alexcrichton:
I don't really have a chance to dig into this right now, but I don't think this is a fix we'd want to merge, it looks like it makes client ergonomics much worse?

tamird:
@alexcrichton is it much worse? it doesn't seem that bad to me. The `#[no_link]` bit is optional, so it's just the matter of asking users to `extern crate core;`.

I don't sufficiently understand the issue in https://github.com/rust-lang/rust/pull/44441#issuecomment-328236190 to suggest another solution. I suspect that the issue there is that bitflags linked against stage0's libcore is being linked into stage1 and that's no bueno.

sfackler:
Forcing all of the hundreds of consumers of bitflags to explicitly pull in core so it can be used in rustc's internals does not seem particularly reasonable.

tamird:
@sfackler all consumers were forced to pull in `std` before https://github.com/rust-lang-nursery/bitflags/pull/29 - the change in ergonomics was accidental.

It seems to me that virtually all consumers will already be pulling in either libcore or libstd, and that this change would not be breaking in practice. Is there a way we can test?

alexcrichton:
@sfackler is right here, this simply needs to be solved a different way. When I get time I can try to help dig into the rustc case, but this isn't going to get landed as-is.

tamird:
Another approach I tried unsuccessfully: make this conditional on a cargo feature, roughly:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 5470c98..55a65f6 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,4 @@ travis-ci = { repository = "rust-lang-nursery/bitflags" }
 default = ["example_generated"]
 unstable_testing = []
 example_generated = []
+no_link = []
diff --git a/src/lib.rs b/src/lib.rs
index 1565582..8c862ef 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,17 +220,12 @@
 #![no_std]
 
 #![doc(html_root_url = "https://docs.rs/bitflags/1.0.0")]
-// When compiled for the rustc compiler itself we want to make sure that this is
-// an unstable crate.
-#![cfg_attr(rustbuild, feature(staged_api))]
-#![cfg_attr(rustbuild, unstable(feature = "rustc_private", issue = "27812"))]
 
 #[cfg(test)]
 #[macro_use]
 extern crate std;
 
-// Re-export libstd/libcore using an alias so that the macros can work in no_std
-// crates while remaining compatible with normal crates.
+#[cfg(not(no_link))]
 #[doc(hidden)]
 pub extern crate core as _core;
 
@@ -366,8 +361,16 @@ macro_rules! __impl_bitflags {
             )+
         }
     ) => {
-        impl $crate::_core::fmt::Debug for $BitFlags {
-            fn fmt(&self, f: &mut $crate::_core::fmt::Formatter) -> $crate::_core::fmt::Result {
+        #[cfg(no_link)]
+        #[allow(unused)]
+        use ::core::fmt::{Binary,Debug,Formatter,LowerHex,Octal,Result,UpperHex};
+
+        #[cfg(not(no_link))]
+        #[allow(unused)]
+        use $crate::_core::fmt::{Binary,Debug,Formatter,LowerHex,Octal,Result,UpperHex};
+
+        impl Debug for $BitFlags {
+            fn fmt(&self, f: &mut Formatter) -> Result {
                 // This convoluted approach is to handle #[cfg]-based flag
                 // omission correctly. For example it needs to support:
                 //
@@ -413,24 +416,24 @@ macro_rules! __impl_bitflags {
                 Ok(())
             }
         }
-        impl $crate::_core::fmt::Binary for $BitFlags {
-            fn fmt(&self, f: &mut $crate::_core::fmt::Formatter) -> $crate::_core::fmt::Result {
-                $crate::_core::fmt::Binary::fmt(&self.bits, f)
+        impl Binary for $BitFlags {
+            fn fmt(&self, f: &mut Formatter) -> Result {
+                Binary::fmt(&self.bits, f)
             }
         }
-        impl $crate::_core::fmt::Octal for $BitFlags {
-            fn fmt(&self, f: &mut $crate::_core::fmt::Formatter) -> $crate::_core::fmt::Result {
-                $crate::_core::fmt::Octal::fmt(&self.bits, f)
+        impl Octal for $BitFlags {
+            fn fmt(&self, f: &mut Formatter) -> Result {
+                Octal::fmt(&self.bits, f)
             }
         }
-        impl $crate::_core::fmt::LowerHex for $BitFlags {
-            fn fmt(&self, f: &mut $crate::_core::fmt::Formatter) -> $crate::_core::fmt::Result {
-                $crate::_core::fmt::LowerHex::fmt(&self.bits, f)
+        impl LowerHex for $BitFlags {
+            fn fmt(&self, f: &mut Formatter) -> Result {
+                LowerHex::fmt(&self.bits, f)
             }
         }
-        impl $crate::_core::fmt::UpperHex for $BitFlags {
-            fn fmt(&self, f: &mut $crate::_core::fmt::Formatter) -> $crate::_core::fmt::Result {
-                $crate::_core::fmt::UpperHex::fmt(&self.bits, f)
+        impl UpperHex for $BitFlags {
+            fn fmt(&self, f: &mut Formatter) -> Result {
+                UpperHex::fmt(&self.bits, f)
             }
         }
 

```

but the `use` declarations appear not to be hygienic, resulting in errors like:
```
error[E0252]: the name `UpperHex` is defined multiple times
    --> src/lib.rs:370:79
     |
370  |           use $crate::_core::fmt::{Binary,Debug,Formatter,LowerHex,Octal,Result,UpperHex};
     |                                                                                 --------
     |                                                                                 |
     |                                                                                 `UpperHex` reimported here
     |                                                                                 previous import of the trait `UpperHex` here
...
1031 | /         bitflags! {
1032 | |             struct PrivateFlags: i8 {
1033 | |                 const FLAG_Y = 0;
1034 | |             }
1035 | |         }
     | |_________- in this macro invocation
     |
     = note: `UpperHex` must be defined only once in the type namespace of this module
```

Anyway, would love some help, here.

PR #122, opened 2017-09-27
Change all the examples to avoid double-namespacing

Amanieu:
Since bitflags now use associated constants, there is no longer any need to prefix individual flags with `FLAG_`. This PR updates the documentation to reflect this.


alexcrichton:
👍

PR #127, opened 2017-10-17
Move tests to test suite crate

### LINKED ISSUE #68: Create a test suite crate
#61 adds an `unstable_testing` feature that is only considered by tests. It seems unfortunate to leak this into the public API.

Having a separate crate for tests ([like in Serde](https://github.com/serde-rs/serde/tree/master/test_suite)) would avoid this.

tamird:
Closes #68.

Also, removed travis-cargo now that docs are handled by docs.rs and
don't need to be uploaded from travis.

Minor update to the compiletest script to avoid issues with stale
artifacts left around after `rustup update`.

r? @dtolnay


tamird:
This is green now - @dtolnay could you have a look?

Issue #130, opened 2017-10-22
Support building without core


dtolnay:
As described in https://github.com/rust-lang/rust/pull/45079#issuecomment-336722610 bitflags currently depends on core [here](https://github.com/rust-lang-nursery/bitflags/blob/862582d107bb74ce4c7b505b2490eb815bc3a8c2/src/lib.rs#L231), but as a macro-only crate it shouldn't need to. This makes it impossible to use bitflags within the implementation of libcore.

dtolnay:
@tamird may be interested in working on this.

tamird:
Yep, was planning to work on that after https://github.com/rust-lang-nursery/bitflags/pull/127 merged. Could you give that a review?

tamird:
OK, I pulled on this thread a bit, but it seems that `path` variables can't be chained. In other words, we'd need a separate path variable for every one of the identifiers being brought in, not just one for `core`. Am I missing something?

dtolnay:
Is this what you mean by can't be chained?

```rust
macro_rules! m {
    ($p:path) => {
        use $p::HashMap;
        //    ^^ expected one of `;` or `as` here
    };
}

m!(std::collections);
```

If so, something like this might work as a workaround.

```rust
macro_rules! m {
    ($($p:ident)::+) => {
        use $($p)::+::HashMap;
    };
}
```

tamird:
That's exactly what I meant, thank you! I'll try that.

dtolnay:
We are not planning to pursue this. https://github.com/rust-lang-nursery/bitflags/pull/134#issuecomment-338702315

PR #134, opened 2017-10-22
[WIP] Support building without core

tamird:
@dtolnay this is my initial work toward #130, which currently doesn't compile. I'd appreciate any guidance you have!

```
$ cargo test
   Compiling bitflags v1.0.0 (file:///Users/tamird/src/bitflags)
error: expected one of `!`, `+`, `for`, `where`, or `{`, found `::`
   --> src/lib.rs:385:19
    |
385 |           impl $Core::fmt::Debug for $BitFlags {
    |                     ^^ expected one of `!`, `+`, `for`, `where`, or `{` here
    |
   ::: src/example_generated.rs
    |
4   | / bitflags! {
5   | |     /// This is the same `Flags` struct defined in the [crate level example](../index.html#example).
6   | |     /// Note that this struct is just for documentation purposes only, it must not be used outside
7   | |     /// this crate.
...   |
15  | |     }
16  | | }
    | |_- in this macro invocation

error: Could not compile `bitflags`.
warning: build failed, waiting for other jobs to finish...
error: expected one of `!`, `+`, `for`, `where`, or `{`, found `::`
   --> src/lib.rs:385:19
    |
385 |           impl $Core::fmt::Debug for $BitFlags {
    |                     ^^ expected one of `!`, `+`, `for`, `where`, or `{` here
    |
   ::: src/example_generated.rs
    |
4   | / bitflags! {
5   | |     /// This is the same `Flags` struct defined in the [crate level example](../index.html#example).
6   | |     /// Note that this struct is just for documentation purposes only, it must not be used outside
7   | |     /// this crate.
...   |
15  | |     }
16  | | }
    | |_- in this macro invocation

error: Could not compile `bitflags`.

To learn more, run the command again with --verbose.
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -361,7 +361,7 @@ macro_rules! __bitflags {
         }
 
         __impl_bitflags! {
-            $BitFlags: $T {
+            $BitFlags: $T: $crate::_core {
                 $(
                     $(#[$inner $($args)*])*
                     $Flag = $value;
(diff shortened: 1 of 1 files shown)


alexcrichton:
I do not personally think we should merge this. No Rust code should be built without libcore as libcore is at the base of everything. We may one day wish to add runtime support to this crate and we wouldn't be able to do so otherwise. I don't think there's much to gain by using this literal macro as-is in libcore.

dtolnay:
I agree with Alex's comment. Thanks anyway!

PR #136, opened 2017-11-12
Adding missing #[inline]

Amanieu:
There are a few places that are missing `#[inline]`. This causes `all()` to produce inefficient code at runtime when the bitflags are defined in a separate crate.


dtolnay:
I published this in 1.0.1.

Amanieu:
Thanks!

PR #149, opened 2018-03-02
Fix clippy lints

hmvp:
The docs are not yet published for the latest clippy but this pull request fixes two clippy warnings you get when using `bitflags!` in your code

https://rust-lang-nursery.github.io/rust-clippy/v0.0.187/index.html#redundant_field_names
https://rust-lang-nursery.github.io/rust-clippy/v0.0.187/index.html#suspicious_arithmetic_impl


niklasf:
Regarding the implementation of Sub we should consider making clippy a bit more relaxed: https://github.com/rust-lang-nursery/rust-clippy/issues/2510

niklasf:
Update: Clippy has been adjusted #2511, so we no longer need to ignore `suspicious_arithmetic_impl`.

hmvp:
Updated the change to remove the now unneeded allow

mulkieran:
It looks like clippy may have made both allows unneeded now: https://github.com/stratis-storage/devicemapper-rs/pull/287

niklasf:
Yes, redundant field names are ignored in macros now: https://github.com/rust-lang-nursery/rust-clippy/commit/ed769a3bc4b39ebca5e4f5b9d8449c1cb8aa0cff.

We should still commit this small cleanup. It's not a breaking change because the shorthand syntax was stabilized in Rust 1.17 and bitflags already requires 1.20.

PR #150, opened 2018-03-06
Make clippy happy: redundant_field_names

niklasf:
This is not a breaking change because the shorthand syntax was [stabilized in Rust 1.17](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1170-2017-04-27) and bitflags already requires 1.20.


niklasf:
Sorry, duplicate of #149.

Issue #28, opened 2016-01-02
Add ability to iterate through flags


gsingh93:
Iterating over each each possible flag value would be a cool feature. I'm currently making an array with each constant value and then iterating over that.

RReverser:
@alexcrichton Is this something that would be accepted in the library?

alexcrichton:
Seems plausible!

djg:
I was just thinking about implementing this feature for some code I'm working on, if it's not already implemented.

Dushistov:
May be generate const array or const function that return array that contains
all values?

Then implementation of iteration will be strait forward,
plus the array may also contains names:

```Rust
const fn all_values() -> [(Flag, &'static str); N]
```

This helps in implementation of custom serialization/deserialization (#147) or custom debug (#168).

What do you think?

At no I want parse / generate strings like: "AB BC CD" <-> `Flags::Ab | Flags::Bc | Flags::Cd`,

to generate string I need iteration plus names of constants in upper-case,
to deserialize I need names in upper-case

sam0x17:
Would love this

KodrAus:
In #278 we merged an implementation that allows iterating over flags and their names. It effectively looks like:

```rust
pub fn iter(self) -> impl Iterator<Item = (&'static str, Self)>
```

It effectively does what @Dushistov described, but internally. There's still design work to be done on this though, `IntoIterator` to implement, and some refactoring of how we generate flags types so we can make that `iter` function return a concrete type instead of `impl Trait` so I'll keep this ticket open for now to consider those.

KodrAus:
This is now implemented in `2.0.0-rc.1`

PR #152, opened 2018-03-29
Add support for iterating set bits in flags.

djg:
This implements #28.

This is gated by feature = "iterator" because it adds a new trait and struct, as well as using `num-traits` crate to aid the generic implementation of `struct BitFlagsIter`.

I'm looking for feedback on the implementation of `BitFlagsIter`. I tried to implement this with out introducting the `BitFlags` trait, but I couldn't work out how to create a `bitflags!` struct with out the trait to enable conversion from raw bits to wrapped type.

I also tried implementing `fn next()` with out the help from `num-traits` but I had trouble getting `Wrapping` for `Sub` to resolve.

This needs documentation before it can land.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,9 @@ travis-ci = { repository = "rust-lang-nursery/bitflags" }
 [features]
 default = ["example_generated"]
 example_generated = []
+iterator = ["num-traits"]
+
+[dependencies.num-traits]
+version = "0.2"
+default-features = false
+optional = true
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -1,0 +1,67 @@
+// Copyright 2014 The Rust Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use num_traits::int::PrimInt;
+use num_traits::ops::wrapping::WrappingSub;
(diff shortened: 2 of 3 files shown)


RReverser:
Maybe instead of separate `BitFlagsIter` implement iterator directly on generated bitflags structure or a simple newtype around it? Technically all you need for iteration is the current mask where at each time one significant bit is taken, emitted and cleared from the mask, so iterating over bitmask (which already implements `Copy`) would do just that.

djg:
@RReverser Thanks for your feedback. I'll have a look at simplifying.

With the lack of `concat_idents!` in macros I chose to implement the newtype wrapper via the generic `BitFlagsIter`, but then ran into issues trying to work on flags generically since there is currently no trait implemented which can be used to constrain `T`.

djg:
FYI: I'm still interested in this, I just haven't had time to get the addition to work as I'd hoped.

Palladinium:
What work needs to still be done on this? This is a feature I'd like to see as well, and I'd be willing to finish it off to see it happen.

Dushistov:
@Palladinium 

> What work needs to still be done on this?
> and I'd be willing to finish it off to see it happen

Stabilization of `concat_idents` will solve of problems:

https://github.com/rust-lang/rust/issues/29599

So it would be possible to introduce new type via:

```Rust
struct concat_idents!($BitFlags, Iterator) {
```

Or if you can resolve problem with `const fn` inside traits: https://github.com/rust-lang/rust/issues/57563 it would be possible to add something like

```Rust
trait BigFlagsAllValues<T> {
   const fn values() -> &'static [T];
}
impl BitFlagsAllValues for $BitFlags 
```

clouds56:
According to the thread, `concat_idents` could be only used in places accept `expr` and things like
```rust
struct concat_idents!($BitFlags, Iterator) {
```
would never compiles.

Shall we look for some other solution?
I think the `Iterator` trait is so fundamental for bitflags.

KodrAus:
In #278 we merged an implementation of `Iterator` for set flags and their names.

Thanks for originally pushing this along @djg!

PR #153, opened 2018-04-09
Fix the example `Flags` struct

hcpl:
So that it actually match the one in the crate level docs.


dtolnay:
I put back the `pub` in 0f42a0085df182cc5a2dbf15a4a97711eca33fac.

PR #156, opened 2018-04-26
Don't include example off docs.rs

CAD97:
This uses [docs.rs metadata](https://github.com/onur/docs.rs/pull/73) to specify the inclusion of the example, meaning that it doesn't have to be included by default.

Technically a [breaking-change], though... (though the module is specified as not public API)


dtolnay:
Published in 1.0.2. This drops compile time of the bitflags crate by itself from 0.41 seconds to 0.29 seconds on my machine, a 30% improvement. :grinning:

PR #157, opened 2018-04-30
Improve zero value flag handling and documentation.

### LINKED ISSUE #151: Bitflags values of zero are considered always present
When a bitflag is built with a zero valued item then it is treated as always present by bitflags functions.

This may cause bugs but it definitely causes confusion when printing such a value. See the following for an example:
```
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Flags: u32 {
        const NONE = 0b0;
        const SOME = 0b1;
    }
}

fn main() {
    let none = Flags::NONE;
    let some = Flags::SOME;
    
    println!("NONE = {:?}", none);
    println!("SOME = {:?}", some);
    assert!(some.contains(Flags::NONE));
}
```
Results in an output of:
```
NONE = NONE
SOME = NONE | SOME
```

I noticed this when printing some flags from git2, specifically [MergeAnalysis](https://docs.rs/git2/0.7.1/git2/struct.MergeAnalysis.html). The output of a `format("{:?}", flag)` ended up being "(ANALYSIS_NONE | ANALYSIS_NORMAL | ANALYSIS_FASTFORWARD)" which confused me greatly. ANALYSIS_NONE is zero as that is how it is defined in [git2](https://github.com/libgit2/libgit2/blob/HEAD/include/git2/merge.h#L320)

It seems to me that way that flag is intended to be used is that NONE only shows up when there are no other things to set, or maybe as a placeholder and should only appear when nothing else is set.

Alex-Addy:
PR to resolve #151.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,6 +215,36 @@
 //!     assert_eq!(implemented_default, (Flags::A | Flags::C));
 //! }
 //! ```
+//!
+//! # Zero Flags
+//!
+//! Flags with a value equal to zero will have some strange behavior that one should be aware of.
+//!
+//! ```
+//! #[macro_use]
+//! extern crate bitflags;
+//!
+//! bitflags! {
+//!     struct Flags: u32 {
+//!         const NONE = 0b00000000;
+//!         const SOME = 0b00000001;
+//!     }
+//! }
+//!
+//! fn main() {
+//!     let empty = Flags::empty();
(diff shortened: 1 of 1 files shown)


dtolnay:
I released 1.0.3 with this fix.

Alex-Addy:
Fantastic, I'm happy to be able to give back to such a useful library.

PR #158, opened 2018-05-15
Add a feature for use with nightly that allows marking BitFlag::empty as a const fn

alex:
Not sure if there's interest in having this, but I'd find it useful since it'd let me set `empty()` as an associated const on a `trait`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -508,8 +508,15 @@ macro_rules! __impl_bitflags {
             )+
 
             /// Returns an empty set of flags.
+            #[cfg(not(feature = "const_empty"))]
             #[inline]
             pub fn empty() -> $BitFlags {
+                $BitFlags { bits: 0 }
+            }
+
+            #[cfg(feature = "const_empty")]
+            #[inline]
+            pub const fn empty() -> $BitFlags {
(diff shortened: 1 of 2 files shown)


alex:
I also came across a use for this with `bits()`. If there's interest in this problem I can add that to this PR.

src/lib.rs
```diff
@@ -508,11 +508,18 @@ macro_rules! __impl_bitflags {
             )+
 
             /// Returns an empty set of flags.
+            #[cfg(not(feature = "const_empty"))]
             #[inline]
             pub fn empty() -> $BitFlags {
                 $BitFlags { bits: 0 }
             }
 
+            #[cfg(feature = "const_empty")]
```
dtolnay: I think this does something other than what you intended. This attribute ends up in the output of the bitflags macro so it would look for a `const_empty` feature in the caller's crate, not the `const_empty` feature you added above.

alex: _Sigh_. You're right. This is easily worked around by defining the method in your own codebase. In light of that, I think I'm just going to close this, and send a PR once `const fn` is stabilized.

PR #163, opened 2018-06-04
Update issue templates

Dylan-DPC:


dtolnay:
I am mildly :-1: on this template. I think making the process of reporting an issue feel institutional discourages casual reporting of minor issues, and I would always rather know about minor issues than have every issue be immaculately detailed. If we use issue templates their focus should be on encouraging contribution and not on milking possibly irrelevant details out of every user.

Dylan-DPC:
I understand your view. As a person opening issues, I just feel a detailed approach is better since the author knows what details to provide. It is 4-5 questions and doesn't feel too taxing. We have something similar set up for uuid crate.

dtolnay:
The uuid templates colored my opinion of this. The most recent issue I reported to uuid I came very close to not reporting at all because neither the feature request template nor the bug report template were really appropriate and I didn't feel like dealing with them.

Dylan-DPC:
@dtolnay yeah just realised when I was creating this issue that I didn't remove the earlier template (which was in place before github allowed you to create separate templates directly instead of adding a parameter in the url). Fixing that right now :smile:

KodrAus:
Perhaps the template would be the most useful if we frame it as a suggested way to compose an issue, rather than like a mandated form? A sentence at the very top like

> This template has some example sections to help you describe your issue. Feel free to fill in or remove them as you need.

with the current set of questions could be enough to help folks who'd like a structure to follow get started but invite others to structure their issues however they feel is most appropriate?

I guess the problem with that though is that we'll probably end up with that text in opened issues unless we worked it into the example text of each section.

At least the new issue template flow in GitHub makes it easy to drop back to regular issues if you don't think the template is suitable.

Dylan-DPC:
It is also so that we get enough details from the user. People might forget to mention some of these details if it isn't there in the template (as they wouldn't know what we are expecting them to tell us)

KodrAus:
Personally, I'm not too worried about not having enough detail up front, because lots of feature requests evolve as conversation takes place.

What if we shelve this one for now, see how the templates using the new GitHub flow in `uuid` work and then revisit it down the track?

Dylan-DPC:
@KodrAus yah i'm fine with that.

PR #171, opened 2019-01-23
Fix running tests on on "stable" Rust

Mart-Bogdan:
Currently all tests can be run only on nightly toolchain.
This PR fixes this, by using compiletest_rs with feature "stable" which works on stable rust, as well as on Nightly.

**P.S.** I wanted to make some use of cfg flags, but this solution seems working, despite loading but more packages then with nightly only.

**P.P.S.** What about specifying exact version of **compiletest_rs**, like "0.3.18", which is current, to not break further builds/tests ?


Mart-Bogdan:
Update: now executing `--features unstable` tests in Travis on every configuration except 1.20 as it's incompatible with lazy_static required by compiletest_rs

```
$ cargo tree --features unstable -ip lazy_static
lazy_static v1.2.0
└── thread_local v0.3.6
    └── regex v1.1.0
        └── compiletest_rs v0.3.18
            └── test_suite v0.0.0 (C:\Users\martb\CLionProjects\bitflags\test_suite)

```

KodrAus:
Thanks @Mart-Bogdan! This looks good to me. Let's go and change that `*` version of `compiletest_rs` to whatever the latest is :+1:

Mart-Bogdan:
Good, @KodrAus I've fixed version.

I've cloned code, was intending to implement serde_json serialization, but noticed that all tests aren't working on stable :-)

But actually I've found macros overwhelming, I was expecting it to be procedural ones.
I'm new to Rust, but perhaps will do some more **PR**s soon.

Mart-Bogdan:
>For `serde` serialization, we should think about how that would work first before diving in

Yep, I've seen existing Issue, and big problem is that it already serializes, using nested object, so this should be compatible :-)

PR #172, opened 2019-03-11
const functions

### LINKED ISSUE #170: Make `bits()` method const
Now that simple const functions are stable in Rust, it'd be useful to make the `bits()` method const. I've just added a wrapper around the `.bits` field to do this for a particular use case (assigned a bit-or of flags to another constant that needs to be of the underlying type) - but it'd be nice if the bitflags macro could do it.

arturoc:
This adds const functions for every non mutable function in the crate.

It also adds a new join function so bitwise or of flags can be done as a const function as well

To avoid the crate from stop working with older versions of the compiler it detects the version using a build.rs script as advised in https://github.com/bitflags/bitflags/issues/170

Closes #170

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,44 @@
+use std::env;
+use std::process::Command;
+use std::str::{self, FromStr};
+
+fn main(){
+    let minor = match rustc_minor_version() {
+        Some(minor) => minor,
+        None => return,
+    };
+
+    // const fn stabilized in Rust 1.31:
+    if minor >= 31 {
+        println!("cargo:rustc-cfg=const_fn");
+    }
+}
+
+fn rustc_minor_version() -> Option<u32> {
+    let rustc = match env::var_os("RUSTC") {
+        Some(rustc) => rustc,
+        None => return None,
+    };
+
+    let output = match Command::new(rustc).arg("--version").output() {
+        Ok(output) => output,
+        Err(_) => return None,
+    };
+
+    let version = match str::from_utf8(&output.stdout) {
+        Ok(version) => version,
+        Err(_) => return None,
+    };
+
+    let mut pieces = version.split('.');
+    if pieces.next() != Some("rustc 1") {
+        return None;
+    }
+
+    let next = match pieces.next() {
+        Some(next) => next,
+        None => return None,
+    };
+
+    u32::from_str(next).ok()
+}
--- a/build.rs
+++ b/build.rs
@@ -10,7 +10,7 @@ fn main(){
 
     // const fn stabilized in Rust 1.31:
     if minor >= 31 {
-        println!("cargo:rustc-cfg=const_fn");
+        println!("cargo:rustc-cfg=bitflags_const_fn");
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -418,7 +418,7 @@ macro_rules! __bitflags {
 #[macro_export(local_inner_macros)]
 #[doc(hidden)]
 #[cfg(const_fn)]
-macro_rules! __fn {
+macro_rules! __fn_bitflags {
     (
         $(#[$filtered:meta])*
         fn $($item:tt)*
@@ -438,7 +438,7 @@ macro_rules! __fn {
 #[macro_export(local_inner_macros)]
 #[doc(hidden)]
 #[cfg(not(const_fn))]
-macro_rules! _fn {
+macro_rules! __fn_bitflags {
     (
         $(#[$filtered:meta])*
         fn $($item:tt)*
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -769,6 +833,63 @@ macro_rules! __impl_bitflags {
         $(#[$filtered])*
         fn $($item)*
     };
+
```
KodrAus: nit: there are a few extra newlines here.

src/lib.rs
```diff
@@ -415,6 +415,46 @@ macro_rules! __bitflags {
     };
 }
 
+#[macro_export(local_inner_macros)]
+#[doc(hidden)]
+#[cfg(const_fn)]
+macro_rules! __fn_bitflags {
+    (
+        $(#[$filtered:meta])*
+        fn $($item:tt)*
```
KodrAus: At some point I think we'll find ourselves bumping our minimum supported rustc version past `1.31.0` and won't need this anymore. So it might make sense just to require the `const` keyword in the input tokens for both `__fn_bitflags` macro impls even though older compilers won't be generating `const` functions:

```rust
__fn_bitflags! {
    #[inline]
    pub const fn bits(&self) -> $T {
        self.bits
    }
}
```

src/lib.rs
```diff
@@ -556,33 +602,51 @@ macro_rules! __impl_bitflags {
...
-                (*self & other) == other
+            __fn_bitflags! {
+                #[inline]
+                pub fn contains(&self, other: $BitFlags) -> bool {
+                    (self.bits & other.bits) == other.bits
+                }
+            }
+
+            /// Joins to set of flags into a new one
+            __fn_bitflags! {
```
KodrAus: Shall we pull this new function out for now so we can review just the new machinery for `const` functions?

build.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn main(){
+    let minor = match rustc_minor_version() {
+        Some(minor) => minor,
+        None => return,
+    };
+
+    // const fn stabilized in Rust 1.31:
+    if minor >= 31 {
+        println!("cargo:rustc-cfg=const_fn");
```
KodrAus: Following the [example of `regex`](https://github.com/rust-lang/regex/blob/master/build.rs#L16-L29) I think we should call this `bitflags_const_fn` so it's namespaced just for `bitflags` and clear that it's a cfg we're setting.

src/lib.rs
arturoc: This is fixed now

src/lib.rs
arturoc: I've removed this function by now. I can send it in a different PR though, without it the const functionality is not very useful since you can only check 1 flag at a time

arturoc:
@KodrAus I've fixed the problems in the review.

Also there's one non mutable function that can't be turned into `const` yet (i think): `from_bits` since it contains a conditional. I've tried to find a way to map a `bool` to the Option required for that function but i don't think there's a way without using an if or a match which are still not supported for `const`functions

arturoc:
Also before merging this, be aware that right now rls seems to crash when using this branch. I've opened an issue in the rust repo: https://github.com/rust-lang/rust/issues/59134#issuecomment-472389137  but it's not solved yet

KodrAus:
Ok, it looks like we've regressed our docs on generated bitflags with the introduction of the `__fn_bitflags` macro:

Screen Shot 2019-04-09 at 1 40 18 pm

We'll probably need to shift the doc comments into the `__fn_bitflags` macro itself for them to return.

arturoc:
I believe the docs regression is solved now

src/lib.rs
```diff
@@ -420,17 +420,17 @@ macro_rules! __bitflags {
 #[cfg(bitflags_const_fn)]
 macro_rules! __fn_bitflags {
     (
-        $(#[$filtered:meta])*
+        $(#[$attr:ident $($attr_args:tt)*])*
```
RReverser: These could be simplified to `$(# $attrs:tt)*`, since the macro doesn't care about internal structure and only needs to propagate tokens.

src/lib.rs
```diff
@@ -600,41 +600,41 @@ macro_rules! __impl_bitflags {
...
+                /// Returns `true` if all flags are currently set.
                 #[inline]
                 pub const fn is_all(&self) -> bool {
                     self.bits == $BitFlags::all().bits
                 }
             }
 
-            /// Returns `true` if there are flags common to both `self` and `other`.
             __fn_bitflags! {
+            /// Returns `true` if there are flags common to both `self` and `other`.
```
RReverser: Indentation seems odd.

arturoc:
Both are fixed now

Dushistov:
Hello, @KodrAus !

Is something blocked this PR from merging?

KodrAus:
bors retry

KodrAus:
Hmm, looks like bors is unhappy about something so isn’t able to merge the PR. I’ll dig into it and see if I can figure it out.

KodrAus:
@arturoc are you happy to squash these commits down and rebase on the current `master`? If that doesn't make bors happy then I'll just merge this one in manually.

PR #174, opened 2019-04-01
Add function to inclusively return bits that do not correspond to a flag

jabedude:
Hello, this PR would add a function that would allow all bits to be returned, even if they do not correspond to a flag. This would be useful in cases where flags may be added to a given bit field without a wrapping library being updated to reflect the new flags.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -473,6 +473,14 @@ macro_rules! __impl_bitflags {
                         f.write_str(__bitflags_stringify!($Flag))?;
                     }
                 )+
+                if self.extra_bits() != 0 {
+                    if !first {
+                        f.write_str(" | ")?;
+                    }
+                    first = false;
+                    f.write_str("0x")?;
+                    $crate::_core::fmt::LowerHex::fmt(&self.extra_bits(),f)?;
+                }
                 if first {
                     f.write_str("(empty)")?;
                 }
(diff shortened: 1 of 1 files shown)


KodrAus:
Hi @jabedude :wave: Thanks for the PR. I'm a little hesitant to introduce a hole in the `bitflags` API that makes it possible to get a hold of an instance of a bitflags type that isn't actually valid though.

Is there a concrete usecase you've got where you find yourself wanting this?

jabedude:
Hi @KodrAus! My use case is using the `nix` crate, specifically the `fcntl` API. That crate wraps a number of known flags in a bitflag bitfield for getting a given file descriptors status flags. There is no robust way to return all flags after a F_GETFL call if say Linux adds flags and the `nix` crate does also add them.

KodrAus:
Hmm, if forwards compatibility is a concern for the `nix` crate then I think that's better solved there by adding inherent methods on their flags types to allow building flags while retaining those extra bits. The `extra_bits` method itself is a bit of a compatibility hazard though if you find yourself relying on it because when, for example, the `nix` crate caught up with newly added variants you'd stop receiving the same extra bits you were previously relying on.

KodrAus:
I'll close this one in favour of #188 

Thanks again @jabedude

PR #175, opened 2019-04-23
Const fn

arturoc:
This adds const functions for every non mutable function in the crate.


arturoc:
Trying to squash and rebase against master https://github.com/bitflags/bitflags/pull/172

KodrAus:
Let's see how this goes!

bors r+

arturoc:
yay!

PR #184, opened 2019-07-01
add from_bits_[ref|mut] for FFI in place ops

flier:
When use FFI struct, we always need to copy the field to a `flags` and write it back after change it
```rust
#[repr(C)]
struct Foo {
   bar: u32,
}

let mut flags = Flags::from_bits_truncate(foo.bar);
// ...
flags |= Flags::SOME;
// ...
foo.bar = flags.bits();
```
With `from_bits_ref` and `from_bits_mut` method, we could support in place operations, like
```rust
let flags = Flags::from_bits_mut(&mut foo.bar);
// ...
*flags |= Flags::SOME;
```
Every change to the underlying field could be read immediately,
and the mutable reference can be pass/return to other function, no more write back.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ travis-ci = { repository = "bitflags/bitflags" }
 
 [features]
 default = []
+unsafe = []
 example_generated = []
 
 [package.metadata.docs.rs]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,6 +613,7 @@ macro_rules! __impl_bitflags {
             /// A cheap reference-to-reference conversion.
             /// Used to convert from an existing bit representation,
             /// unless that representation contains bits that do not correspond to a flag.
+            #[cfg(feature = "unsafe")]
             #[inline]
             fn from_bits_ref(bits: &$T) -> Option<&$BitFlags> {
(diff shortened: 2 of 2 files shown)


KodrAus:
Thanks for the PR @flier!

I'm a little concerned about introducing `unsafe` into the macros, because it will break folks compiling with `#[deny(unsafe)]`, and leave no option for folks compiling with `#[forbid(unsafe)]`.

The `#[repr(transparent)]` change looks like a good one though.

flier:
I think we may introduce an `unsafe` feature for those scenes which need work with `unsafe` FFI code, and disable it by default for compatibility. @KodrAus

KodrAus:
@flier Hmm, it would be unfortunate to leak the fact that this requires `unsafe` to implement that way. It's not a discoverable way to flag it. So I think our best path forward in `bitflags` is to add just the `#[repr(transparent)]` attribute so it's valid for you to write the cast in your own FFI code.

We could add a section under our crate docs called _Layout_ or _FFI_, and note that your bitflags has the same size and alignment of `T`, but not all valid bit patterns of `T` are guaranteed to be valid flags.

flier:
Fair, I have submit another PR #187 to add the `#[repr(transparent)]` attribute, thank anyway

Issue #137, opened 2017-11-16
Constants cannot be cfg-gated multiple times


nox:

sfackler:
Why not? This seems to work just fine:

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Flags: u32 {
        #[cfg(target_os = "linux")]
        const FOO = 1;
    }
}

fn main() {
    let _ = Flags::FOO;
}
```

nox:
@sfackler Thanks for the answer, I just realised I misformulated this issue.

This doesn't work:

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Flags: u32 {
        #[cfg(target_os = "linux")]
        const FOO = 1;
        #[cfg(not(target_os = "linux"))]
        const FOO = 2;
    }
}

fn main() {
    let _ = Flags::FOO;
}
```

I've worked around it doing this:

```rust
bitflags! {
    struct Flags: u32 {
        const FOO = {
            #[cfg(target_os = "linux")] { 1 }
            #[cfg(not(target_os = "linux"))] { 2 }
        };
    }
}
```

sfackler:
Weird!

KodrAus:
It looks like the problem here is our logic that attempts to handle flags that have been cfg'd out. We intentionally omit attributes from flags to ensure they're known to the debug format and the `all` function. Unfortunately this doesn't handle flags where the same flag is defined with different values based on cfg. I haven't quite unpicked _why_ we need the names of all flags to be available, but I think the way we currently do it might just be plain broken given a flag isn't just either present or not, it may have a totally different value.

KodrAus:
It looks like it might actually be a bit of an artefact from pre-associated-const days (how far we’ve come). It might be possible to simplify these now and either avoid the need for those traits or just apply the cfgs to them.

PR #186, opened 2019-07-20
fix up failing test and ensure it runs in CI

### LINKED ISSUE #185: `cargo build --test` fails on Windows
[One of the tests](https://github.com/bitflags/bitflags/blob/8a10bdc144bb2a4b97c63cfa76be90228954d029/src/lib.rs#L920) for `bitflags-1.1.0` fails to compile on Windows with a stable rustc (version 1.36.0).

```
[INFO] [stderr] error[E0425]: cannot find value `_CFG_A` in this scope
[INFO] [stderr]    --> src\lib.rs:927:28
[INFO] [stderr]     |
[INFO] [stderr] 927 |             const _CFG_C = _CFG_A.bits | 0b10;
[INFO] [stderr]     |                            ^^^^^^ not found in this scope
[INFO] [stderr] 
[INFO] [stderr] error: aborting due to previous error
```

This suggests that `#[cfg]` attributes are not being properly added to `const` declarations, but I'm not quite sure exactly what's going on here.

KodrAus:
Closes #185 

We have a failing test on Windows that's been going unnoticed for a while, because we don't test Windows in CI.

This isn't a bug in `bitflags` itself, just one of its tests not being updated.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -919,12 +919,12 @@ mod tests {
 
     bitflags! {
         struct _CfgFlags: u32 {
-            #[cfg(windows)]
-            const _CFG_A = 0b01;
             #[cfg(unix)]
-            const _CFG_B = 0b01;
+            const _CFG_A = 0b01;
             #[cfg(windows)]
-            const _CFG_C = _CFG_A.bits | 0b10;
+            const _CFG_B = 0b01;
+            #[cfg(unix)]
+            const _CFG_C = Self::_CFG_A.bits | 0b10;
         }
     }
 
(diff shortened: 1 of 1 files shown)


ecstatic-morse:
Sorry I'm a bit late to the party. I imagined a proper test would look [like the following](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0cbdd8c385fb2c9cf423dbb06d1c7f6f).

```rust
bitflags! {
    struct Cfg: u8 {
        #[cfg(unix)]
        const CFG_A  = 0b1;
        #[cfg(not(unix))]
        const CFG_A  = 0b10;
        
        const CFG_B  = Self::CFG_A.bits & 0b110;
    }
}
```

However, this fails to build. Is this expected?

ecstatic-morse:
Doh, this is just #137.

PR #187, opened 2019-07-26
add #[repr(transparent)]

flier:
add `#[repr(transparent)]` for the the FFI layout, replace #184.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -418,6 +418,7 @@ macro_rules! __bitflags {
         }
     ) => {
         $(#[$outer])*
+        #[repr(transparent)]
         #[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
         $($vis)* struct $BitFlags {
             bits: $T,
@@ -588,7 +589,7 @@ macro_rules! __impl_bitflags {
                         )*
                     }
                 }
-            }            
+            }
 
             /// Returns the raw value of the flags currently stored.
             #[inline]
@@ -1018,12 +1019,12 @@ mod tests {
             unsafe { Flags::from_bits_unchecked(0b1001) },
             (extra | Flags::A)
         );
-      
+
         let extra = unsafe { EmptyFlags::from_bits_unchecked(0b1000) };
         assert_eq!(
-          unsafe { EmptyFlags::from_bits_unchecked(0b1000) },
-          (extra | EmptyFlags::empty())
-      );
+            unsafe { EmptyFlags::from_bits_unchecked(0b1000) },
+            (extra | EmptyFlags::empty())
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for the PR @flier!

It looks like we'll have to bump our minimum supported Rust version to `1.28.0` to support the representation. That release is about 12 months old now so I think this is a reasonable change.

josephlr:
I think this is a reasonable change and it would help me avoid boilerplate w/ FFI stuff.

I don't think an MSRV bump is needed, we can just check the version in the build.rs [like we do for `const fn` stuff](https://github.com/bitflags/bitflags/blob/15e911c304d5bd8805af55d7e4f8c5324ed798ee/build.rs#L11).

@KodrAus does that sound reasonable? @flier I'm happy to take this over if you don't have time.

flier:
Hi @josephlr 

  I have added a `bitflags_repr_transparent` feature gate for the  `#[repr(transparent)] ` attribute, could you help me to review the changes? 
  Thanks

elmarco:
Hi @flier, can you update the MR? thanks

KodrAus:
Thanks for your patience on this one @flier. Wow I really let this sit for a long time! 1.28.0 is now _really_ quite old so I think we should be comfortable just bumping to it if we haven't already and avoid needing to add feature gates.

I was wondering whether `#[repr(transparent)]` came with a bit pattern guarantee (in `bitflags` we currently enforce that all set bits correspond to valid flags, so not all bit patterns of the underlying integer are necessarily valid bitflags), but since non-zero types like `NonZeroUsize` are also `#[repr(transparent)]` I think we're safe here.

If you'd like to update this PR to just unconditionally add the `#[repr(transparent)]` attribute or if somebody else would like to add that we can merge this in.

KodrAus:
In #240 we bumped all the way up to `1.45.0`, so can definitely unconditionally add the `#[repr(transparent)]` flag now 🙂

KodrAus (approved):
Thanks @flier! This looks good to me.

elmarco:
thanks! release time? :)

njaard:
This breaks at least the crates nix and rusqlite, because they both use #[repr(C)] on "enum structs"

mockersf:
also wgpu-types and naga, which already added `#[repr(transparent)]`
```
error[E0692]: transparent struct cannot have other repr hints
    --> /.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-types-0.7.0/src/lib.rs:2791:12
     |
2779 | / bitflags::bitflags! {
2780 | |     /// Flags for which pipeline data should be recorded.
2781 | |     ///
2782 | |     /// The amount of values written when resolved depends
...    |
2791 | |     #[repr(transparent)]
     | |            ^^^^^^^^^^^
...    |
2812 | |     }
2813 | | }
     | |_^
     |
     = note: this error originates in the macro `bitflags::bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)
```

KodrAus:
Thanks for the reports everyone :+1: I've gone ahead and yanked `1.3.0` so we can fix this up. I really wish we had something like `crater` we could use to test things like this. But in the meantime I'll add some compilation tests that include other `repr`s

Issue #191, opened 2019-09-27
Could not compile on Rasbian Buster with flag --release


rbolog:
Hi,
I got an error on Rasberry pi with Raspbian Buster only when I use flag --release of cargo . Otherwise it works fine. Have you any advice ? 
Note: I have no problem with linux x86_64 and windows.
Thanks

> 'A macro to generate structures which behave like bitflags.
> ' CARGO_PKG_HOMEPAGE='https://github.com/bitflags/bitflags' CARGO_PKG_VERSION_PATCH=0 CARGO_MANIFEST_DIR=/home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/bitflags-1.2.0 CARGO_PKG_AUTHORS='The Rust Project Developers' CARGO=/home/pi/.rustup/toolchains/stable-armv7-unknown-linux-gnueabihf/bin/cargo CARGO_PKG_REPOSITORY='https://github.com/bitflags/bitflags' rustc --crate-name build_script_build /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/bitflags-1.2.0/build.rs --color always --crate-type bin --emit=dep-info,link -C opt-level=3 --cfg 'feature="default"' -C metadata=b8f67689abe13aac -C extra-filename=-b8f67689abe13aac --out-dir /home/pi/dev/sha3sum/target/release/build/bitflags-b8f67689abe13aac -L dependency=/home/pi/dev/sha3sum/target/release/deps --cap-lints warn`
> error: Could not compile `bitflags`.

A

rbolog:
It's an issue with version 1.38
see: https://github.com/rust-lang/rust/issues/62896/ 
The workaround is:
[profile.release]
codegen-units = 1
Thanks

PR #192, opened 2019-09-28
Check build on a no-std target

KodrAus:
For #191 

Make sure we actually compile against a `no-std` target and can generate `bitflags` definitions.


KodrAus:
The build failure is unrelated. It looks like we're having trouble compiling `compiletest_rs` for the integration test suite.

PR #194, opened 2019-10-11
Remove `#[inline]` on consts

varkor:
`#[inline]` has no effect on consts. It looks like these attributes were added when they were functions, and was never removed.


varkor:
@KodrAus: would it be possible to publish a new minor version with this change? Thanks!

PR #202, opened 2019-12-03
End `empty` doc comment with full stop

glts:
Trivial: `empty` is the only function whose generated API doc doesn’t end with a full stop. Let’s add one.


Dylan-DPC:
Thanks

PR #203, opened 2019-12-10
Implement Iterator for bitflags

arturoc:
Allows to use a bitflags type as an iterator, iterating through all the enabled flags and consuming the original in the process.

This implements `Iterator` using `trailing_zeros` and also `ExactSizeIterator` using count_ones.


arturoc:
just tried with some crates that use this one and it produces some errors. going to close this and try something else instead

PR #205, opened 2019-12-16
Modernization and default generation

npmccallum:
This PR contains five patches. Four do modernization work. One implements a new `default()` suffix feature.

I pursued modernization first because it simplified the macro structure before adding a new macro feature.

My main interest is the `default()` suffix feature. The existing methods for doing `impl Default` have served us fairly well. However, their main drawback is that we don't have the opportunity to generate a `const fn default()` for use in constant contexts.


KodrAus:
Thanks for the PR @npmccallum!

I think we should consider the modernization effort independently of the new `default` suffix feature. For the modernization we should be careful not to regress any existing callers.

For the `default` suffix feature we should spend some time exploring the design space and see what approaches we could take. Let's open up a separate issue for that.

konsumlamm:
@npmccallum do you plan to continue working on this? If not, I'd try to extract the modernization part and open a new PR for that.

npmccallum:
@konsumlamm  go ahead!

Issue #200, opened 2019-11-03
Document why from_bits_unchecked is unsafe


little-dude:
Hello,

From the documentation `from_bits_unchecked()` looks like what I need:

> Convert from underlying bit representation, preserving all bits (even those not corresponding to a defined flag).

However the method is `unsafe`, so I'd like to know under which conditions I can safely use it, and what are the risks if I don't use it correctly.

KodrAus:
Hi @little-dude 👋 

We sketched out the reasoning for making `from_bits_unchecked` unsafe [in the original PR](https://github.com/bitflags/bitflags/pull/188#issuecomment-517586059). The reason is that you become responsible for enforcing the flags have a “valid” bit pattern, which is usually enforced by picking some combination of the flags that were supplied by whoever defined the flags type originally. With `from_bits_unchecked` we don’t guarantee what’s in the flags is constrained to patterns allowed by that list. Does that make sense?

That should definitely make its way into the docs in some form!

little-dude:
Hi @KodrAus,

Thank you! The thread you pointed answers my question. I understand the reason for marking this as `unsafe`. I find it a little weird (or maybe _unconventional_ is a better term) to use `unsafe` to warn about potential _logic_ mistakes, instead of memory management mistakes, so I think it worth making it clear in the documentation indeed. I'll try to send a PR this week-end.

Thanks for your time!

PR #207, opened 2020-01-22
Document from_bits_unchecked unsafety (#200)

niklasf:
Every unsafe method should have documented safety invariants. Here's an attempt for `from_bits_unchecked()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -637,6 +637,15 @@ macro_rules! __impl_bitflags {
             __fn_bitflags! {
                 /// Convert from underlying bit representation, preserving all
                 /// bits (even those not corresponding to a defined flag).
+                ///
+                /// # Safety
+                ///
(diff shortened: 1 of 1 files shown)


SOF3:
How exactly can the `bitflags! {}` caller decide whether to allow extra flags? Furthermore, how exactly could unchecked bits actually lead to *memory unsafety* (rather than just unexpected behaviour)?

KodrAus:
@SOF3 Whether or not to allow extra flags is determined by the source-of-truth for those flags. The motivation for this feature was for `bitflags!` that mirror another source that might change from under them. It's unsafe in the same sense that transmuting an integer to boolean is unsafe. Not every possible bit-pattern is a valid value for flags, so as the caller you're guaranteeing the bits you provide are "valid", where the definition of valid depends on the source of the flags.

PR #211, opened 2020-02-04
Let is_all() ignore extra bits (fixes #208)

### LINKED ISSUE #208: is_all() vs. from_bits_unchecked()
[`unsafe from_bits_unchecked()`](https://docs.rs/bitflags/1.2.1/bitflags/example_generated/struct.Flags.html#method.from_bits_unchecked) allows creating instances with extra bits. The caller of the `bitflags!` macro can decide if this is allowed for their type. Let's assume it is for `Example`. I checked the provided methods for surprising interactions with extra bits, and found (only) this:

`is_all()` returns **false** when there are *more* than "all" flags. This does not match the documentation:

> Returns true if all flags are currently set.

Should we update the documentation or the implementation?

---

```rust
use bitflags::bitflags;

bitflags! {
    struct Example: u32 {
        const A = 1;
    }
}

fn main() {
    unsafe {
        assert!(Example::from_bits_unchecked(1).is_all()); // true
        assert!(Example::from_bits_unchecked(3).is_all()); // false
    }
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cda2672387dd0ff4ba629b1317a9c57c

niklasf:
Changes `is_all()` to match the documented behavior in the presence of extra bits:
> Returns `true` if all flags are currently set.

This is only relevant when `unsafe from_bits_unchecked()` has been used to create an instance with bits that do not correspond to any defined flag.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -664,7 +664,7 @@ macro_rules! __impl_bitflags {
                 /// Returns `true` if all flags are currently set.
                 #[inline]
                 pub const fn is_all(&self) -> bool {
-                    self.bits == $BitFlags::all().bits
+                    $BitFlags::all().bits | self.bits == self.bits
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1048,6 +1048,11 @@ mod tests {
         assert!(!Flags::A.is_all());
         assert!(Flags::ABC.is_all());
 
+        let extra = unsafe { Flags::from_bits_unchecked(0b1000) };
+        assert!(!extra.is_all());
+        assert!(!(Flags::A | extra).is_all());
+        assert!((Flags::ABC | extra).is_all());
```
niklasf: This was previously false, even though all flags (and one more bit) were set.

KodrAus:
This looks reasonable to me! Thanks @niklasf

PR #212, opened 2020-02-12
Allow multiple bitflags in one macro invocation

konsumlamm:
Add support for multiple bitflags in one macro invocation, for example:
```rust
bitflags! {
    pub struct Flags1: u32 {
        const A = 0b00000001;
    }

    struct Flags2: u32 {
        const B = 0b00000010;
    }
}
```
The changes to the macro are backwards compatible, so I changed the version to "1.3.0".
I also updated the documentation and replaced mentions of the `flags` keyword (which was used before 0.9) with `struct`.


konsumlamm:
Note that the Ci is only failing, because the expected error message for `test_suite/tests/compile-fail/private_flags.rs` is different on nightly. That is not caused by this commit. See https://github.com/bitflags/bitflags/issues/213.

konsumlamm:
Any opinion on this?

konsumlamm:
Closed in favor of #235.

PR #216, opened 2020-05-07
Fix changed error message

ldm0:


ldm0:
Green already, closed.

PR #217, opened 2020-05-07
Constant operator overloading for bitflags

### LINKED ISSUE #180: Can't use generated bitflags to initialise a const
This works in plain Rust

```rust
const MAMMAL: u8 = 1;
const AQUATIC: u8 = 2;
const SEAL: u8 = MAMMAL | AQUATIC;
```

but the equivalent when using bitflags does not

```rust
use bitflags::bitflags;

bitflags! {
    pub struct FLAGS: u8 {
        const MAMMAL  = 1;
        const AQUATIC = 2;
    }
}

const SECOND_SEAL: FLAGS = FLAGS::MAMMAL | FLAGS::AQUATIC;
```

You get an error *error[E0015]: calls in constants are limited to constant functions, tuple structs and tuple variants* when trying to BITOR the two flags.

It would be nice if it did, because in my program I have a large initial array of flags to create and there are many different combinations making it impractical to spell them all out inside the `bitflags!` call.

Edit: Simplified the repro.

### LINKED ISSUE #210: Using bitflags in static structs
I need to develop a plugin for a program, which requires me to export a certain structure which contains a bitfield. When exporting the struct with `bitflags` I get the following error: 

```cmd
error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants
  --> src/lib.rs:18:12
   |
18 |     flags: PluginFlags::A | PluginFlags::C,
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

I would prefer to use this crate over plain constants... Is there a way to solve this?

code to reproduce:
```rust
use bitflags::bitflags; // 1.2.1

bitflags!{
    pub struct PluginFlags: u32 {
        const None = 0;
        const A = 0x0001;
        const B = 0x0002;
        const C = 0x0004;
    }
}

pub struct Plugin {
    flags: PluginFlags,
}

#[no_mangle]
static DATA_EXPORT: Plugin = Plugin {
    flags: PluginFlags::A | PluginFlags::C,
};
```

ldm0:
Fixes #180 Fixes #210 
After <https://github.com/rust-lang/rust/pull/68847>, we now can overload the operator as const function. This PR added a feature `"constant"`. And a user can write compile-pass code like this:
```rust
#![feature(const_fn)]
#![allow(incomplete_features)]
#![feature(const_trait_impl)]
use bitflags::bitflags;

bitflags! {
    struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
        const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
    }
}

const B: Flags = Flags::A | Flags::B;

fn main() {
    println!("{:?}", B);
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ travis-ci = { repository = "bitflags/bitflags" }
 [features]
 default = []
 example_generated = []
+constant = []
 
 [package.metadata.docs.rs]
 features = [ "example_generated" ]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,6 +475,7 @@ macro_rules! __fn_bitflags {
 
 #[macro_export(local_inner_macros)]
 #[doc(hidden)]
+#[cfg(not(feature = "constant"))]
 macro_rules! __impl_bitflags {
     (
         $BitFlags:ident: $T:ty {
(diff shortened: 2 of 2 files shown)


KodrAus:
Thanks for the PR @ldm0! It looks like things are steadily moving forwards with const trait impls, so I think we'd be best waiting for that to land in some complete form first to save us too much infrastructure to manage in the `bitflags` library itself.

js2xxx:
I think this feature should move forward provided that the `const_fn` feature has been stable since long.

KodrAus:
Thanks for spiking this out @ldm0! We're still waiting on some kind of `const` overloading support in Rust itself, but I think that will be the best time for us to take advantage of it in `bitflags`. I'll go ahead and close this one for now since we won't be doing conditional support for unstable features in this library.

KodrAus:
Thanks for spiking this out @ldm0! We're still waiting on some kind of `const` overloading support in Rust itself, but I think that will be the best time for us to take advantage of it in `bitflags`. I'll go ahead and close this one for now since we won't be doing conditional support for unstable features in this library.

PR #220, opened 2020-06-18
BitFlags trait

### LINKED ISSUE #154: Add BitFlags trait implemented by every bit flags
It would allow to use any bitflags as generic for example:

```rs
struct SomeStruct<F: BitFlags>{
....
}

impl<F: BitFlags> SomeStruct{
    fn set_flags(&mut self, flags: F){
        ...
    }
}
```

it would also allow to document the crate without need for the example_generated module

I can send a PR implementing this if there's interest

arturoc:
Adds a BitFlags trait that is implemented by every bitflags.

The trait includes documentation so it the current hack to document methods through an example could be removed

Closes #154

### DIFF
--- a/src/bitflags_trait.rs
+++ b/src/bitflags_trait.rs
@@ -1,0 +1,54 @@
+pub trait BitFlags<T>:
+    core::ops::BitOr<Output = Self>
+    + core::ops::BitOrAssign
+    + core::ops::BitXor<Output = Self>
+    + core::ops::BitXorAssign
+    + core::ops::BitAnd<Output = Self>
+    + core::ops::BitAndAssign
+    + core::ops::Sub<Output = Self>
+    + core::ops::SubAssign
+    + core::ops::Not<Output = Self>
+    + core::iter::Extend<Self>
+    + core::iter::FromIterator<Self>
+{
+    /// Returns an empty set of flags.
+    fn empty() -> Self;
+    /// Returns the set containing all flags.
+    fn all() -> Self;
+    /// Returns the raw value of the flags currently stored.
+    fn bits(&self) -> T;
+    /// Convert from underlying bit representation, unless that
+    /// representation contains bits that do not correspond to a flag.
+    fn from_bits(bits: T) -> core::option::Option<Self>;
+    /// Convert from underlying bit representation, dropping any bits
+    /// that do not correspond to flags.
+    unsafe fn from_bits_truncate(bits: T) -> Self;
+    /// Convert from underlying bit representation, preserving all
+    /// bits (even those not corresponding to a defined flag).
+    ///
+    /// # Safety
+    ///
+    /// The caller of the `bitflags!` macro can chose to allow or
+    /// disallow extra bits for their bitflags type.
+    ///
+    /// The caller of `from_bits_unchecked()` has to ensure that
+    /// all bits correspond to a defined flag or that extra bits
+    /// are valid for this bitflags type.
+    unsafe fn from_bits_unchecked(bits: T) -> Self;
+    /// Returns `true` if no flags are currently stored.
+    fn is_empty(&self) -> bool;
+    /// Returns `true` if all flags are currently set.
+    fn is_all(&self) -> bool;
+    /// Returns `true` if there are flags common to both `self` and `other`.
+    fn intersects(&self, other: Self) -> bool;
+    /// Returns `true` all of the flags in `other` are contained within `self`.
+    fn contains(&self, other: Self) -> bool;
+    /// Inserts the specified flags in-place.
+    fn insert(&mut self, other: Self);
+    /// Removes the specified flags in-place.
+    fn remove(&mut self, other: Self);
+    /// Toggles the specified flags in-place.
+    fn toggle(&mut self, other: Self);
+    /// Inserts or removes the specified flags depending on the passed value.
+    fn set(&mut self, other: Self, value: bool);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -262,6 +262,11 @@ extern crate std;
 #[doc(hidden)]
 pub extern crate core as _core;
 
+#[doc(inlined)]
+pub use bitflags_trait::BitFlags;
+
+mod bitflags_trait;
+
 /// The macro used to generate the flag structure.
 ///
 /// See the [crate level docs](../bitflags/index.html) for complete documentation.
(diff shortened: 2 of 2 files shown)


Adam-Gleave:
Any updates on this? I've been looking for something like this for a project, and the changes look reasonable.

src/bitflags_trait.rs
```diff
@@ -0,0 +1,68 @@
+use core as _core;
+
+pub trait BitFlags<T>:
+    _core::ops::BitOr<Output = Self>
```
KodrAus, line 4: I think it might be better for the `BitFlags` trait not to include these additional supertraits and let consumers add them as additional bounds if they need them.

On the one hand it seems reasonable that if you've got flags you'll want to perform bitwise operations on them, but on the other it's a bit of a grab-bag snapshot of our current functionality that we won't be able to extend (since we can't add new supertraits without breakage).

It might be best to stick with just the current set of methods on the trait which cover:

- Construction from bits
- Conversion back into bits
- Set operations on pairs of flags

That should be enough to support a wide range of fundamental use-cases.

src/bitflags_trait.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// Returns the set containing all flags.
+    fn all() -> Self;
+    /// Returns the raw value of the flags currently stored.
+    fn bits(&self) -> T;
+    /// Convert from underlying bit representation, unless that
+    /// representation contains bits that do not correspond to a flag.
+    fn from_bits(bits: T) -> _core::option::Option<Self>;
+    /// Convert from underlying bit representation, dropping any bits
+    /// that do not correspond to flags.
+    unsafe fn from_bits_truncate(bits: T) -> Self;
```
KodrAus, line 39: This method doesn't need to be `unsafe`.

KodrAus:
Thanks for your patience on this one everybody! I've had a look through this now and left some thoughts on what direction we might want to take it in :bow:

src/bitflags_trait.rs
```diff
@@ -0,0 +1,68 @@
+use core as _core;
+
+pub trait BitFlags<T>:
```
KodrAus, line 3: We should probably make this an associated type, since an implementation of `BitFlags` only corresponds to a single underlying integer type.

src/bitflags_trait.rs
arturoc: What about Clone + Copy? I need those to implement the iterator which is based on a value and removes the flags as it advances through it.

src/bitflags_trait.rs
arturoc: I think the lines in this comment are wrong? do you mean perhaps empty and all?

KodrAus: Ah sorry, I mean making the trait look like:

```rust
pub trait BitFlags {
    type Bits;

    ..
}
```

instead of:

```rust
pub trait BitFlags<T> {
    ..
}
```

src/bitflags_trait.rs
KodrAus: Do we need the `BitFlags` type itself to be `Copy`? Or just the underlying bits? If I understand right, we could just perform shifts and masks on the bits themselves and use `BitFlags::from_bits_unchecked` to return just that flag.

arturoc:
I've implemented all the changes and left out Copy and Clone by now since they are not really needed for this PR, once this is merged I'll look into what's the minimum needed for the iterators PR.

Two of the tests are failing but doesn't seem related to this, the 1.32 test seems to be related to unsupported features in const function because that function is not wrapped with `__fn_bitflags!` I'll send another PR to fix that.  The nightly tests are failing cause the output format has changed so the comparison with the generated private_flags.stderr is not equal anymore

bradleyharden:
@arturoc, we ran into a case where this would be useful for the [atsamd-hal](https://github.com/atsamd-rs/atsamd) crate. I'm looking forward to seeing it get merged. Thanks for working on it.

KodrAus:
Hey @bradleyharden!  :wave:

Would you be happy to share some more details of your usecase? Would the trait in its current shape without any supertraits be sufficient for you?

KodrAus:
Looks like we just need a rebase on the current `main`, which should fix up those build failures.

Thanks again for persisting with this @arturoc

src/bitflags_trait.rs
```diff
@@ -0,0 +1,46 @@
+use core as _core;
```
konsumlamm, line 1: This is only necessary for use in macros, so that types get resolved correctly. For the definition of the trait, you can just write `core::option::Option` etc.

konsumlamm:
We may also want to add the functions added in #244 as trait methods (which would also eliminate the need for the supertraits for ops). I was worried a bit about the names conflicting with the respective `const fn`s (since trait methods can't be `const` currently, sadly), but there seem to be no problems.

KodrAus:
>  I was worried a bit about the names conflicting with the respective const fns (since trait methods can't be const currently, sadly), but there seem to be no problems.

Ah yes, inherent methods are resolved over methods from traits or deref, so we'd be safe adding them to the trait.

We may also want to use some private supertrait to discourage callers from implementing this trait directly (we can't technically prevent it since expansion happens in their crates). That way we could mark the trait methods as `const` in the future when it's possible without needing a major version bump. Maybe something like:

```rust
#[doc(hidden)]
pub mod _private {
    pub trait ProducedByMacro {}
}

pub trait BitFlags: ProducedByMacro { .. }
```

konsumlamm:
> We may also want to use some private supertrait to discourage callers from implementing this trait directly (we can't technically prevent it since expansion happens in their crates). That way we could mark the trait methods as `const` in the future when it's possible without needing a major version bump. Maybe something like:
> 
> ```rust
> #[doc(hidden)]
> pub mod _private {
>     pub trait ProducedByMacro {}
> }
> 
> pub trait BitFlags: ProducedByMacro { .. }
> ```

Why would this private supertrait have to be generated by a macro? It could just be defined as a normal trait (as is the case for the `BitFlags` trait in this PR). That way, we would have a truly sealed trait.

src/bitflags_trait.rs
```diff
@@ -0,0 +1,46 @@
...
+    type Bits;
+    /// Returns an empty set of flags.
+    fn empty() -> Self;
+    /// Returns the set containing all flags.
+    fn all() -> Self;
+    /// Returns the raw value of the flags currently stored.
+    fn bits(&self) -> Self::Bits;
+    /// Convert from underlying bit representation, unless that
+    /// representation contains bits that do not correspond to a flag.
+    fn from_bits(bits: Self::Bits) -> _core::option::Option<Self>
```
konsumlamm, line 13:
```suggestion
    fn from_bits(bits: Self::Bits) -> Option<Self>
```

src/bitflags_trait.rs
```diff
@@ -0,0 +1,46 @@
+use core as _core;
+
+pub trait BitFlags {
```
konsumlamm, line 3:
```suggestion
/// A trait that is automatically implemented for all bitflags.
pub trait BitFlags {
```

KodrAus:
> Why would this private supertrait have to be generated by a macro? It could just be defined as a normal trait

Ah the pattern would be almost the same as the usual `private::Sealed` approach. I just pulled the name `ProducedByMacro` out to suggest to anyone looking at the docs that our requirement on implementing `ProducedByMacro` is that it came from the `bitflags!` macro.

The sealed trait (whether it's called `ProducedByMacro`, `Sealed`, or anything else) would live in `bitflags`, but unlike in the usual pattern we'd still need it to be publicly exported so that callers of `bitflags!` in external crates could still implement it along with `BitFlags`. We just don't want anyone implementing that sealed trait directly. And if they do they run the risk of breakage when we tweak `BitFlags`.

arturoc:
jjust found a moment to rebase on current main, every test seems to be passing now without problem

PR #227, opened 2020-10-13
Making crate accessible from std

raoulstrackx:
Changing dependencies to enable the crate to be used from inside the rust standard library.

see https://github.com/rust-lang/rust/tree/master/library/rustc-std-workspace-core

cc: @jethrogb

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,17 @@ exclude = [
 ]
 build = "build.rs"
 
+[dependencies]
+core = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-core' }
+compiler_builtins = { version = '0.1.2', optional = true }
+
 [badges]
 travis-ci = { repository = "bitflags/bitflags" }
 
 [features]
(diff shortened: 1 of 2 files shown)


raoulstrackx:
@alexcrichton could this get a review?
All build jobs pass, except for the rust 1.20 compiler.That one fails with `no matching package named `core` found (required by `bitflags`)`. Presumably that compiler is too old to make it compatible to include it in std. Note that that compiler is from 2017-08-27, so over 3 years old.

alexcrichton:
Sorry but I do not maintain this crate.

jethrogb:
You're still listed as a crate owner on crates.io

KodrAus (approved):
This looks good to me!

raoulstrackx:
Thanks for merging! Could you also publish the changes to crates.io? Thanks!

raoulstrackx:
@KodrAus Could you please publish these changes on crates.io as well? Much appreciated!

PR #229, opened 2020-10-19
Make from_bits a const fn

bprosnitz:
'if' now works in 'const fn as of rust 1.4.6 (https://blog.rust-lang.org/2020/08/27/Rust-1.46.0.html).

I would love to have 'from_bits' be a 'const fn' because it is the safest from_* function.


KodrAus:
This looks good to me!

@bprosnitz would you like to rebase this on `master` to bring in some CI fixes?

PR #231, opened 2020-11-06
Migrate to github actions

KodrAus:
Fixing up our current CI woes so that we can get back on track with `bitflags`!


ratijas:
Hi. What about `bors.toml` file? There is still a mention of travis CI.

```toml
status = [
	"continuous-integration/travis-ci/push",
]
```

There is some [documentation](https://bors.tech/documentation/getting-started/) about this status checks. But I don't understand it very well.

PR #235, opened 2020-12-27
Allow multiple bitflags structs in one macro invocation

konsumlamm:
Add support for multiple bitflags structs in one macro invocation, for example:

```rust
bitflags! {
    pub struct Flags1: u32 {
        const A = 0b00000001;
    }

    struct Flags2: u32 {
        const B = 0b00000010;
    }
}
```

I also updated the documentation, replaced mentions of the `flags` keyword (which was used before 0.9) with `struct` and ran `cargo fmt`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::process::Command;
 use std::str::{self, FromStr};
 
-fn main(){
+fn main() {
     let minor = match rustc_minor_version() {
         Some(minor) => minor,
         None => return,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,7 @@
 //! A typesafe bitmask flag generator useful for sets of C-style bitmask flags.
 //! It can be used for creating typesafe wrappers around C APIs.
 //!
-//! The `bitflags!` macro generates a `struct` that manages a set of flags. The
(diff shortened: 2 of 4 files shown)


KodrAus:
Thanks @konsumlamm! This looks like a nice quality-of-life improvement to me. I can't quite spot from the diff, do we still have any tests that create a single flags type? That seems like a case worth keeping.

konsumlamm:
> I can't quite spot from the diff, do we still have any tests that create a single flags type? That seems like a case worth keeping.

Yes, there are still plenty tests that only create a single flags type, for example in [line 1379](https://github.com/konsumlamm/bitflags/blob/multiple-bitflags/src/lib.rs#L1379) or [line 1444](https://github.com/konsumlamm/bitflags/blob/multiple-bitflags/src/lib.rs#L1444) and the tests in `test_suite` all still use single flags types. It may be worth to add a test for the empty case (`bitflags! {}`), but that's just [one line](https://github.com/konsumlamm/bitflags/blob/multiple-bitflags/src/lib.rs#L405) and there's not much there that could go wrong.

konsumlamm:
@KodrAus ?

KodrAus:
Thanks for your patience on this one @konsumlamm. I had a bit of a break from things.

This looks good to me!

PR #238, opened 2021-05-09
Remove remaining mentions of Travis CI

ratijas:
Disclaimer: I have no idea what am I doing!

image

***

Badges section has been removed from crates.io; it is suggested to place
them in README instead[1], which we already have done.

In bors.toml, the status section. Its is unclear from the docs[2], what those
strings are. But it seems to me, they should reference name of GitHub
Workflow file with CI pipelines[3].

[1]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section
[2]: https://bors.tech/documentation/
[3]: https://bors.tech/devdocs/bors-ng/readme.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ description = """
 A macro to generate structures which behave like bitflags.
 """
 exclude = [
-    ".travis.yml",
     "appveyor.yml",
     "bors.toml"
 ]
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks @ratijas! Let's let this run through and see what bors thinks of it 🙂

PR #242, opened 2021-05-17
Fixex compilation for rustc 1.32 or less

arturoc:
The function `from_bits` was declared as const but not wrapped in the `__bitflags_fn` macro. The `__bitflags_fn` turns const into non const functions for rustc 1.31 or higher but this function contains conditionals which aren't supported until 1.33

This PR adds a new macro and cfg flag in build.rs which turns anything wrapped in `__bitflags_fn_cond` from const to non const fn if the rustc version is 1.32 or less.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,8 +13,8 @@ fn main() {
         println!("cargo:rustc-cfg=bitflags_const_fn");
     }
 
-    // const fn conditionals stabilized in Rust 1.32:
-    if minor >= 32 {
+    // const fn conditionals stabilized in Rust 1.33:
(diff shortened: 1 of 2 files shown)


konsumlamm:
Since #240 the minimum supported Rust version is 1.46, so I don't think this is necessary anymore.

KodrAus:
Thanks for the PR @arturoc!

In general, I've ended up prioritising keeping the macro implementation simple over compatibility with older compilers (within reason). As @konsumlamm mentioned, since we've bumped our MSRV for 1.46 we don't need to explicitly support anything below that.

PR #244, opened 2021-06-10
Add named functions to perform set operations

thomcc:
There are a few reasons to do this:

1. Tooling and documentation is much better about exposing inherent methods than trait implementations. This is true for rustdoc, but it also is (inherently) the case for autocomplete (I can do `thing.` and see all the methods available for a type, but it won't show me operator overloads)

2. Somewhat subjective, but IMO there are situations where these names are clearer than using bitwise operators. They certainly are much easier to find documentation for, and it's more obvious at a glance that you aren't working with e.g. plain integers.

3. It unblocks the ability to perform these operations to initialize `const`s/`static`s and within `const fn`. I gather that the stance of this repo in the past been to wait for the operator traits to be usable in CTFE contexts, but this seems *extremely* far off from stabilizing (I asked about this on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/146212-t-compiler.2Fconst-eval/topic/Const.20traits.20temperature.20check), and they confirmed that my impression is correct).

    For full disclosure: this is a lot of my motivation here, but I also do genuinely believe what I wrote for 1 and 2 (that is, I'd like these as part of the API, but probably would not have bothered to write the patch if it were not for this).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,6 +164,18 @@
 //! - `toggle`: the specified flags will be inserted if not present, and removed
 //!             if they are.
 //! - `set`: inserts or removes the specified flags depending on the passed value
+//! - `intersection`: returns a new set of flags, containing only the flags present
+//!                   in both `self` and `other` (the argument to the function).
+//! - `union`: returns a new set of flags, containing any flags present in
+//!            either `self` or `other` (the argument to the function).
+//! - `difference`: returns a new set of flags, containing all flags present in
+//!                 `self` without any of the flags present in `other` (the
+//!                 argument to the function).
+//! - `symmetric_difference`: returns a new set of flags, containing all flags
+//!                           present in either `self` or `other` (the argument
+//!                           to the function), but not both.
+//! - `complement`: returns a new set of flags, containing all flags which are
+//!                 not set in `self`, but which are allowed for this type.
 //!
 //! ## Default
 //!
@@ -653,7 +665,6 @@ macro_rules! __impl_bitflags {
             pub const fn contains(&self, other: $BitFlags) -> bool {
                 (self.bits & other.bits) == other.bits
             }
-
             /// Inserts the specified flags in-place.
             #[inline]
             pub fn insert(&mut self, other: $BitFlags) {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -681,6 +690,96 @@ macro_rules! __impl_bitflags {
...
+            /// Returns the intersection between the flags in `self` and
+            /// `other`.
+            ///
+            /// Specifically, the returned set contains only the flags which are
+            /// present in *both* `self` *and* `other`.
+            ///
+            /// This is equivalent to using the `&` operator (e.g.
+            /// [`ops::BitAnd`]), as in `flags & other`.
+            ///
+            /// [`ops::BitAnd`]: https://doc.rust-lang.org/std/ops/trait.BitAnd.html
```
thomcc, line 703: I asked around and there's
1. No great way to link to the implementation of a trait for `Self`.
2. Not really a good way to link to right location of the trait in all cases, especially when cases like `feature = "rustc-dep-of-std"` are considered.

Note that broken links produces a warning (in some more recent versions of rust), and that warning would appear in the users crate, so it seems better to just explicitly give the link, rather than trying to make intra-doc-links figure it out.

src/lib.rs
```diff
@@ -1156,6 +1255,188 @@ mod tests {
...
+        assert_eq!(e1.intersection(e2), Flags::C);
+        assert_eq!(e1.difference(e2), Flags::A | extra);
+        assert_eq!(e2.difference(e1), Flags::B);
+        assert_eq!(e2.complement(), Flags::A);
+        assert_eq!(e1.complement(), Flags::B);
+        assert_eq!(e1.symmetric_difference(e2), Flags::A | Flags::B | extra); // toggle
+    }
+
+    #[test]
+    fn test_set_ops_exhaustive() {
```
thomcc, line 1329: This should also improve our test coverage of the `ops::Foo` implementations a lot, at least in terms of edge cases.

src/lib.rs
```diff
@@ -1156,6 +1255,188 @@ mod tests {
...
+                const D = 0b000001000;
+                const E = 0b000010000;
+                const F = 0b000100000;
+                const G = 0b001000000;
+                // Intentionally no `H`
+                const I = 0b100000000;
+            }
+        }
+        let iter_test_flags =
+            || (0..=0b111_1111_1111).map(|bits| unsafe { Test::from_bits_unchecked(bits) });
```
thomcc, line 1350: This is a lot of test cases, especially given that the test iterates over it twice (nested), but even for a debug build, it completes in under half a second on a relatively underpowered machine. This is fast enough not to be noticeable at all (that is, the tests don't stall on it, waiting for it to complete).

It probably helps that all it's doing is verifying the result of some bitwise operations (but also: computers are fast).

src/lib.rs
```diff
@@ -653,7 +665,6 @@ macro_rules! __impl_bitflags {
             pub const fn contains(&self, other: $BitFlags) -> bool {
                 (self.bits & other.bits) == other.bits
             }
-
```
thomcc, line 656: Ugh. This line had some trailing whitespace on it, and while I was pretty careful to try to keep this diff clean (... y'all kinda need to run rustfmt, at least arguably), I guess this one slipped through the cracks.

LMK if you'd like me to readd the trailing whitespace that used to be here.

thomcc:
It's not really clear to me if this should be considered a breaking change or not... In theory it is, since it's possible people do `impl Blah` after defining `Blah` with `bitflags!`, but I dunno how likely it is that they've defined these functions. AFAICT (From some grep.app sleuthing) I can't find anything, but that doesn't mean it doesn't exist.

I also could do it as an off-by-default feature, which wouldn't *technically* fix the issue, but could make it so that nobody who isn't opting into this has to deal with any kind of breakage.

KodrAus:
Wow thanks for working on this @thomcc! This is a really thoroughly considered addition.

As for whether or not it's a breaking change, we're in a bit of a tricky position here with adding inherent methods to types defined by `bitflags` because even though we define them we don't technically own them. Normally we'd solve this by defining a trait and using that, but that's not an option here because we can't make those trait methods `const`. The off-by-default feature might help us feel a little better, but in practice it wouldn't prevent breakage in crates already defining these methods without opting in to the feature since some other dependency in the graph might do it for them.

In general I think we should avoid adding inherent methods, but for this _specific_ case I think we could justify adding these methods as they could have reasonably been included pre 1.0, have a single obvious definition, and complement inherent methods that already exist. If there are any other methods that would also fit into this category we might want to include them here too. Can you think of any we should consider? I can't off the top of my head.

KodrAus (approved):

KodrAus:
The only other one I can think of is a `const default` based on #205, but that's out-of-scope for this PR.

thomcc:
> Can you think of any we should consider? I can't off the top of my head.

I can't think of anything else, no.

src/lib.rs
```diff
@@ -1156,6 +1257,188 @@ mod tests {
...
+        assert_eq!(bc.complement(), !bc);
+        assert_eq!(bc.complement(), Flags::A);
+        assert_eq!(ac.symmetric_difference(bc), Flags::A.union(Flags::B));
+        assert_eq!(bc.symmetric_difference(ac), Flags::A.union(Flags::B));
+    }
+
+    #[test]
+    fn test_set_ops_const() {
+        // These just test that these compile and don't cause use-site panics
+        // (would be possible if we had some sort of UB), which is enoug
```
konsumlamm, line 1302: Seems like the comment got cut off there.

konsumlamm:
In the future we may want to replace all the `$BitFlags` in the `impl` blocks with `Self`, but that's out of scope for this PR. I can put together a PR for that after this is merged.

KodrAus:
Alrighty, it looks like we're building up some nice content for a release, so I'll merge this one in and we can take stock of what's on `main`.

src/lib.rs
KodrAus: I'll fix this one up on `main` 🙂

thomcc: Totally didn't see this comment, thanks for taking care of it though.

PR #246, opened 2021-06-27
Modernization

konsumlamm:
This PR takes the modernization parts from #205.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,8 +18,7 @@
 //! # Example
 //!
 //! ```
-//! #[macro_use]
-//! extern crate bitflags;
+//! use bitflags::bitflags;
 //!
 //! bitflags! {
 //!     struct Flags: u32 {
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -249,27 +243,17 @@
...
 
-#[cfg(test)]
-#[macro_use]
-extern crate std;
-
-// Re-export libcore using an alias so that the macros can work without
-// requiring `extern crate core` downstream.
-#[doc(hidden)]
-pub extern crate core as _core;
-
```
konsumlamm, line 263: As far as I can tell, this isn't necessary any more. The macros now use `core::` instead of `$crate::_core::`, which seems to work fine, but I'm not entirely sure if it causes problem with macro hygiene.

KodrAus (approved):
Thanks @konsumlamm! This is a great effort 🙇

src/lib.rs
konsumlamm: I just noticed that this breaks if someone defines `mod core {}` when using `bitflags!`, I'll put together a PR to fix that.

PR #247, opened 2021-07-19
Fix regression & simplify tests

konsumlamm:
This fixes a regression introduced in #246 (see https://github.com/bitflags/bitflags/pull/246#discussion_r672290205), by reverting the specific change, along with adding a test case.

I also took the opportunity to move the tests in `test_suite` back into the crate tests (instead of having a workspace with a `test_suite` crate). AFAICT, this happened because it needed extra features that shouldn't leak, but that isn't the case anymore. This makes it easier to run the tests (just `cargo test` in the main directory).

I also removed `appveyor.yml` from the `exclude` list in `Cargo.toml`, I'm not sure why it was there in the first place though.


KodrAus:
Ah nicely spotted! And thanks for improving the integration tests too 🙇

PR #256, opened 2021-08-13
fix: allowed non_snake_case on trait impls

saiintbrisson:
Resolves bitflags#237.


KodrAus (approved):
Thanks @saiintbrisson! This looks good to me

PR #261, opened 2021-10-23
Add a hidden trait to discourage manual impls of BitFlags

KodrAus:
Followup to #220 

cc @konsumlamm @arturoc

This adds a hidden supertrait to our new `BitFlags` trait that suggests you shouldn't implement it manually. Since we only support `BitFlags` through the `bitflags!` macro it means we could safely add new methods and supertraits (like `Add` and friends) to `BitFlags` without breaking consumers.

I've also just been adding to our test suite wherever I can too so there are some other new tests added there too.


KodrAus:
Thanks @konsumlamm and @arturoc! This should be ready for another look now.

PR #266, opened 2021-12-16
Sanitize `Ok`

### LINKED ISSUE #265: The bitflags macro is not sanitary wrt. standard library types and enumerations
The `bitflags` macro, expanded in the prescence of a definition of the type/value `Ok` errors.
Reproduction code:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3fda3e36c7c6a57e0f7a83c84e56df20

Interestingly, the relevant function, the `fmt` function from the Debug impl, does use `::bitflags::_core::fmt::Result`, however, it merely returns the value `Ok(())`.

konsumlamm:
Fixes #265.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -500,7 +500,7 @@ macro_rules! __impl_bitflags {
                 if first {
                     f.write_str("(empty)")?;
                 }
-                Ok(())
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks @konsumlamm!

PR #268, opened 2022-01-02
Fix bug in `Debug` implementation

### LINKED ISSUE #267: Bug: debug pretty-printing unknown flags display 0x0x
main.rs
```rust
use bitflags::bitflags;

bitflags! {
    struct Flags: u8 {
        const TWO = 0x2;
    }
}

fn main() {
    let value = 0b11;
    let flags = unsafe { Flags::from_bits_unchecked(value) };
    println!("{:?}", flags);
    println!("-----------");
    println!("{:#?}", flags);
}
```
will print the following:
```sh
TWO | 0x1
-----------
TWO | 0x0x1
```

the expected output would either be 0x1 for both, or 1, and 0x1 respectively

konsumlamm:
Fixes #267.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -494,8 +494,7 @@ macro_rules! __impl_bitflags {
                         f.write_str(" | ")?;
                     }
                     first = false;
-                    f.write_str("0x")?;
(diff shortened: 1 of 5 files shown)


KodrAus:
Thanks! This looks good to me.

PR #270, opened 2022-01-23
Add example to the README.md file

tiaanl:
Add a simple example to the README.md file to show the capabilities of the crate.  The README is displayed in many places where developers are evaluating its use and having an example is always useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,33 @@ and this to your source code:
 use bitflags::bitflags;
 ```
 
+## Example
+
+Generate a flags structure:
+
+```rust
+use bitflags::bitflags;
+
+// The `bitflags!` macro generates `struct`s that manage a set of flags.
+bitflags! {
+    struct Flags: u32 {
+        const A = 0b00000001;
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks @tiaanl! This looks good to me, but for the README it might be good to put some comments inline with the `bitflags!` invocation and its use to suggest the macro is how you define a flags type, which then generates a bunch of set-like methods for you. What do you think?

tiaanl:
@KodrAus I copied a small comment from the `lib.rs` docs and also grabbed the example code from there to add more use cases.

KodrAus (approved):
Thanks @tiaanl! Sorry for the radio silence on this one. This looks good to me.

tiaanl:
@KodrAus Thanks for the approval.  I'm not sure why all the checks are failing.  I only changed the README 😕 Any suggestions?

KodrAus:
That's just changes to Rust's compiler messages so nothing to worry about here. I've fixed those up so will merge this one in. Thanks again!

PR #276, opened 2022-04-19
from_bits_(truncate) fail with composite flags

### LINKED ISSUE #275: from_bits accepts non existing flags
```rs
    #[test]
    fn test_from_bits_edge_cases() {
        bitflags! {
            struct Flags: u8 {
                const A = 0b00000001;
                const BC = 0b00000110;
            }
        }


        let flags = Flags::from_bits(0b00000100);
        assert!(flags.is_none());
    }
```

Unless I'm missing something this test should pass but it fails cause from_bits accepts flags that are not declared. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6fd4adbddc8b8740cbd35af2306073ca

This is related to this issue in the implementation of iterators in this PR https://github.com/bitflags/bitflags/pull/204#issuecomment-950304444 Using `from_bits` instead of `from_bits_unchecked` should allow to produce any valid flags that are not a combination of other flags but at the moment `from_bits` seems to accept any flag that is included in a combination even if it's not declared. I can try to send a PR.

arturoc:
When a bitflags type contains composite flags like:

```rs
bitflags! {
    struct Flags: u8 {
            const A = 0b00000001;
            const BC = 0b00000110;
    }
}
```

from_bits and from_bits_truncate would not work as expected allowing flags that are not declared.

Fixes #275

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -559,18 +559,41 @@ macro_rules! __impl_bitflags {
             /// representation contains bits that do not correspond to a flag.
             #[inline]
             pub const fn from_bits(bits: $T) -> $crate::_core::option::Option<Self> {
-                if (bits & !Self::all().bits()) == 0 {
-                    $crate::_core::option::Option::Some(Self { bits })
-                } else {
-                    $crate::_core::option::Option::None
+                if bits == 0 {
+                    return Some(Self{ bits })
                 }
+
+                $(
+                    #[allow(unused_doc_comments, unused_attributes)]
+                    $(#[$attr $($args)*])*
+                    if bits & Self::$Flag.bits == Self::$Flag.bits {
+                        return Some(Self{ bits })
+                    }
+                )*
+
+                None
             }
 
             /// Convert from underlying bit representation, dropping any bits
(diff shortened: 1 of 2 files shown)


arturoc:
This one introduced some new errors in the should fail tests, the last commit updates the corresponding .err file

src/lib.rs
```diff
@@ -1797,4 +1820,32 @@ mod tests {
...
+        bitflags! {
+            struct Flags: u8 {
+                const A = 0b00000001;
+                const BC = 0b00000110;
+            }
+        }
+
+
+        let flags = Flags::from_bits(0b00000100);
+        assert!(flags.is_none());
```
KodrAus, line 1835: Would this still return `None` if the input was `0b00000101`?

arturoc:
That was failing, it's now working and have added a test

KodrAus (approved):
Thanks @arturoc! This looks good to me.

We need to keep the paths for `Some` and `None` mangled like they were, but I'll submit a follow-up that does that.

PR #277, opened 2022-04-19
serde tests fail in CI

arturoc:
Cause Serialize and Deserialize derives are used from serde:: instead of serde_derive::

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1789,7 +1789,7 @@ mod tests {
     }
 
     bitflags! {
-        #[derive(serde::Serialize, serde::Deserialize)]
+        #[derive(serde_derive::Serialize, serde_derive::Deserialize)]
         struct SerdeFlags: u32 {
             const A = 1;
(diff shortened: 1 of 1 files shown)


KodrAus:
Hmm, it looks like our tests have bit-rotted a bit here. If you run them locally with `TRYBUILD=overwrite` it should update those expected build failure output files.

arturoc:
I wasn't sure if I should do that. I'll send a different PR with the fix for the trybuild tests and merge it in these ones

KodrAus:
I should write something up in the contributing docs for it 🙂 If you leave it to me I'll take care of it. I think we might do something special so the output is only checked in beta builds.

arturoc:
Yes that's what I was thinking. It's only in beta so not sure that running TRYBBUILD even using a beta version would overwrite the right files

KodrAus:
Thanks @arturoc. I'll merge this one in and rebase my updates to the test output on top.

PR #278, opened 2022-04-19
Iterator over all the enabled options

arturoc:
Adds a new function that returns an iterator over all the enabled flags.

This implementation takes into account combined flags like the ones described in: https://github.com/bitflags/bitflags/pull/204#issuecomment-950304444

It uses an `impl Iterator` as return type because using an specific Iter type implies using const generics which are not supported in earlier versions of rust and adding a const to the trait in order to express the correct dimension of the array as can be seen in the last solution proposed in #204.

Superseeds #204

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -759,8 +759,7 @@ macro_rules! __impl_bitflags {
                     if self.is_empty() || NUM_FLAGS == 0 {
                         None
                     }else{
-                        for pos in start..NUM_FLAGS {
-                            let flag = OPTIONS[pos];
+                        for flag in OPTIONS[start..NUM_FLAGS].iter().copied() {
                             start += 1;
                             if self.contains(flag) {
                                 self.remove(flag);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -730,6 +730,50 @@ macro_rules! __impl_bitflags {
                 Self::from_bits_truncate(!self.bits)
             }
 
+            /// Returns an iterator over all the flags in this set.
+            pub fn iter(mut self) -> impl Iterator<Item = Self> {
+                let mut options = [
+                    $(
+                        #[allow(unused_doc_comments, unused_attributes)]
+                        $(#[$attr $($args)*])*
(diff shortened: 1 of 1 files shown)


KodrAus:
Ah this is an interesting approach 🤔 Thanks for exploring this @arturoc! I'd love to see some test cases that demonstrate how it works, and ideally we'd be able to replace our debug implementation with it (that might require an implementation that yielded `&'static str`s instead of flags though, but we could call that `iter_names` or something.

arturoc:
Ah yes I forgot to bring the tests from the previous implementation, just pushed them in a new commit

arturoc:
Note how in the first test the flags appear in the wrong order due to the optimization to not go through all the options in every iteration. Removing that would show the flags in the correct order but would be slower for bitflags with a lot of options. I can try to do some benchmarks to understand if it's worth

arturoc:
Just figured a better way to optimize the iterator method so it preserves the order of the flags and the options array is a const now.

src/lib.rs
```diff
@@ -1797,4 +1839,51 @@ mod tests {
...
         }
     }
+
+    #[test]
+    fn test_iter() {
+        bitflags! {
+            struct Flags: u32 {
+                const ONE  = 0b001;
+                const TWO  = 0b010;
+                const THREE = 0b100;
```
KodrAus, line 1849: Would it be possible to add a cfg'd-out flag to this as well just to make sure we handle those? Something like:

```rust
#[cfg(windows)]
const FOUR_WIN = 0b1000;
#[cfg(unix)]
const FOUR_UNIX = 0b10000;
```

and then use those matching cfgs to check that we iterate over those flags when they're available, but don't when they're not?

KodrAus:
Thanks for working on this @arturoc! It's a nice simple approach you've ended up with. I like it!

arturoc:
Thanks! have added a test for cfg

KodrAus (approved):
Nicely done! This looks good to me.

PR #279, opened 2022-04-20
Fix beta test output

KodrAus:
cc @arturoc

It looks like I did go as far as making `TRYBUILD=overwrite` work for our custom `.beta` suffixes we use for these compiler output files 🙂 I stubbed out a `CONTRIBUTING` doc for us to collect any more of this trivia that comes up.


arturoc:
great, thanks! I'll merge it into my PRs as soon as all the tests are passing

arturoc:
Ah I think you need this https://github.com/bitflags/bitflags/pull/277 for the tests to pass. Somehow using derive `serde::Serialize` works in some versions but others require `serde_derive::Serialize` which works for all

PR #264, opened 2021-11-02
Wrap generated flags types in a newtype

KodrAus:
Part of #262 

This lets us support:

- #155
- #168
- #218
- #248

It introduces an internal unnamable type as the single field of a `BitFlags` type that holds all the trait implementations. So instead of us having to commit to implementing traits on your generated flags type, we can instead add them to this internal field type and you can choose to derive them on the flags type based on that implementation.

Given an input like:

```rust
bitflags! {
    /// baz
    pub struct Flags: u32 {
        const A = 0b00000001;
        #[doc = "bar"]
        const B = 0b00000010;
        const C = 0b00000100;
        #[doc = "foo"]
        const ABC = Flags::A.bits() | Flags::B.bits() | Flags::C.bits();
    }
}
```

we'll generate code that looks something like:

<details>

```rust
use bitflags::bitflags;
/// baz
pub struct Flags(<Self as ::bitflags::__private::BitFlagsField>::Field);
const _: () = {
    #[repr(transparent)]
    pub struct __Field(u32);
    #[automatically_derived]
    #[allow(unused_qualifications)]
    impl ::core::clone::Clone for __Field {
        #[inline]
        fn clone(&self) -> __Field {
            {
                let _: ::core::clone::AssertParamIsClone<u32>;
                *self
            }
        }
    }
    #[automatically_derived]
    #[allow(unused_qualifications)]
    impl ::core::marker::Copy for __Field {}
    impl ::core::marker::StructuralPartialEq for __Field {}
    #[automatically_derived]
    #[allow(unused_qualifications)]
    impl ::core::cmp::PartialEq for __Field {
        #[inline]
        fn eq(&self, other: &__Field) -> bool {
            match *other {
                __Field(ref __self_1_0) => match *self {
                    __Field(ref __self_0_0) => (*__self_0_0) == (*__self_1_0),
                },
            }
        }
        #[inline]
        fn ne(&self, other: &__Field) -> bool {
            match *other {
                __Field(ref __self_1_0) => match *self {
                    __Field(ref __self_0_0) => (*__self_0_0) != (*__self_1_0),
                },
            }
        }
    }
    impl ::core::marker::StructuralEq for __Field {}
    #[automatically_derived]
    #[allow(unused_qualifications)]
    impl ::core::cmp::Eq for __Field {
        #[inline]
        #[doc(hidden)]
        #[no_coverage]
        fn assert_receiver_is_total_eq(&self) -> () {
            {
[...]

PR #281, opened 2022-05-03
rework the way cfgs are handled

### LINKED ISSUE #137: Constants cannot be cfg-gated multiple times

### LINKED ISSUE #198: Support allow non_upper_case_globals lint.
I'm trying to create a bitflags with CamelCase naming. However adding #[allow(non_upper_case_globals)] everywhere doesn't suppress the lints.

### LINKED ISSUE #215: Debug formatting leads to less desireable output
[Link to rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=%23%5Bmacro_use%5D%0Aextern%20crate%20bitflags%3B%0A%0Abitflags!%20%7B%0A%20%20%20%20struct%20Flags%3A%20u32%20%7B%0A%20%20%20%20%20%20%20%20const%20A%20%3D%200b00000001%3B%0A%20%20%20%20%20%20%20%20const%20B%20%3D%200b00000010%3B%0A%20%20%20%20%20%20%20%20const%20C%20%3D%200b00000100%3B%0A%20%20%20%20%20%20%20%20const%20ABC%20%3D%20Self%3A%3AA.bits%20%7C%20Self%3A%3AB.bits%20%7C%20Self%3A%3AC.bits%3B%0A%20%20%20%20%7D%0A%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20Flags%3A%3AA%20%7C%20Flags%3A%3AB%20%7C%20Flags%3A%3AC%20)%3B%0A%7D)

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
        const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
    }
}

fn main() {
    println!("{:?}", Flags::A | Flags::B | Flags::C );
}
```

prints:

```bash
A | B | C | ABC
```

I find it somewhat less helpful that both the expanded (`A | B | C`) and "compressed" form (`ABC`) are reported...

Is there a reason behind this? Is this considered more correct for some reason?

KodrAus:
Closes #137 
Closes #198 
Closes #215 

The current approach for handling `#[cfg]`'d flags is to define a trait that includes all names unconditionally, and then only set them to values based on actual `#[cfg]`'s. Unfortunately this means we can't support redefining flags (they'd appear twice in the unconditional list). This approach simplifies things by defining an internal `Bits` trait that captures the operations we require of primitive bits types. We now define `all` as `from_bits_truncate` with all bits set.


KodrAus:
Getting back to having a single macro that defines bitflags types should make it much easier for us to do #264 properly. After that I think we're just about ready for the next major release (the first with the `BitFlags` trait).

hillin:
When is this change going to be released?

hillin:
@KodrAus Is there a plan to release with this feature?

Issue #262, opened 2021-10-25
Considerations for 2.0


KodrAus:
Before we actually release a build with #220 I think it’s a good time to consider any breaking changes that could work their way into a `2.0` release. Without much public surface area I think the risk of multiple `bitflags` versions in a dependency graph is low.

Some I had in mind:

- Reconsider making `from_bits_unchecked` unsafe. The idea was to make it unsafe to construct flags using bits that don’t correspond to flags specified. That way an implementor could consider using things like `unreachable_unchecked` in code that matches on bits. In hindsight this just seems like something people work around rather than make use of. Since we generate types in the scope of the `bitflags!` invocation there’s already nothing stopping someone from manually stuffing extra bits in a flags type. So we can’t rely on `from_bits_unchecked` in `bitflags` code either.
- Consider not deriving any traits by default. Ship an optional proc macro `#[bitflags_derive]` that can be used to add them. That way callers can decide on their own equality and debug implementations if they want.
- Consider changing the serialization implementation to use a text format for human-readable formats and raw bits for others.
- Anything else?

KodrAus:
I've started sketching out an alternative implementation for a `2.x` in #264. It allows us to add `#[derive]` support for external libraries like `arbitrary` without potentially breaking callers, but comes with some potential breakage:

```diff
bitflags! {
+   #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
    struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
-       const ABC = Flags::A.bits | Flags::B.bits | Flags::C.bits;
+       const ABC = Flags::A.bits() | Flags::B.bits() | Flags::C.bits();
    }
}
```

If you want to derive `Serialize`, then you'd add a `serde` feature to `bitflags`, then add it. This is different from today, but would be consistent with any other traits we support, like `Arbitrary`.

This gives you a chance to change the way you implement traits like `Debug` and `PartialOrd` in the same way you would for any other type, but does require a stack of `#[derive]`s to be equivalent to `1.x` types.

How do people feel about this? I don't want to create churn that doesn't seem worthwhile, but if it's palatable then it gives us a way to naturally tweak trait implementations. There's a middle-ground here where you would include `#[derive]`s for some of these types, like `Clone`, `Copy`, and `Hash`, and exclude others that may want alternative implementations like comparison.

What do you all think?

Rua:
[...]

PR #282, opened 2022-05-25
Split generated code into two types

### LINKED ISSUE #155: Allow other semantics of `PartialOrd`
Now bitflags merely derives `PartialOrd` with the mechanism built into rustc, which compares the values numerically. But a flags type is also a Boolean lattice, so it has another natural partial order: `x ≤ y` iff `y.contains(x)`. As i can't disable deriving `PartialOrd` i can't use if the other semantics if wanted.

Either option would be fine:
• Let user disable deriving `PartialOrd` (and ergo `Ord`)
• Let user choose the semantics of `PartialOrd` per type

### LINKED ISSUE #168: Add a way to not implement the Debug trait
I'm using `bitflags!()` on a large table of u32 status codes where the high 16-bits are a sequential number (1, 2, 3, 4 etc.) and the lower 16-bits are bit flags. Together these are OR'd together to make the code.

https://github.com/locka99/opcua/blob/master/types/src/status_codes.rs

This all works fine except for the default `Debug` trait implementation which doesn't work properly with sequential values. This is easiest to demonstrate with a cut-down example:

```
#[macro_use]
extern crate bitflags;

bitflags! {
  pub struct StatusCode: u32 {
    const IS_ERROR                = 0x8000_0000;
    //...
    const Good = 0;
    //...
    const BadEncodingError = 0x8006_0000;
    const BadDecodingError = 0x8007_0000;
  }
}

fn main() {
    println!("{:?}", StatusCode::BadDecodingError);
}
```

This outputs `IS_ERROR | BadEncodingError | BadDecodingError` which is inappropriate in my case. I assume it is just doing a bitwise OR to all the values and appending them to the output string. I would like to be able to implement my own Debug trait and not generate the default one.

For example the bitflags!() macro could have a pattern or form which excludes emitting the `Debug` trait, ... `pub struct (nodebug) StatusCode`

Note I can override the `fmt::Display` trait for discrete cases where I need to print the StatusCode but it doesn't help when debug dumping out structs that have StatusCode as a member.

### LINKED ISSUE #218: No way to override trait implementations generated by bitflags! macro
I am trying to implement [FieldAttributes](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.fieldattributes?view=netcore-3.1) from C#, and it uses `FieldAccessMask` and the `&` operator to separate the field visibility (public, private, etc) from the other flags. This also works in Rust code.

The issue is that there is no way to make it use the `&` in this way in the `fmt::Debug` implementation. [Playgroud link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=739cc5f582e615826c90271ec51d6a88).

This problem could be gotten around if there was a way to override the `fmt::Debug` implementation.

### LINKED ISSUE #228: Add an option for checking for valid bits is in the getter rather than constructor
Hi, I want to use bitflags for generated code that interacts with data in other languages, some of which are bitflags (https://github.com/google/flatbuffers/pull/6098). I don't want to drop data so I'd like to use `from_bits_unchecked`. However, it is marked as unsafe. I'd like to submit a PR to add an option to get around this. I have two ideas:

1) Move the valid-bits-check to the getter: There'll be only one constructor: `from_bits` and the `bits()` getter becomes `bits() -> Result<...>`, `unsafe bits_unchecked()`, and `bits_truncated()`

2) #200 documents that this crate uses `unsafe` to mean a usage issue rather than a memory issue, so maybe I could make an option to remove the unsafe altogether, and update the generated documentation appropriately?

What are your thoughts on these ideas? Does bitflags depend on the "bits are truncated to defined flags" invariant?

### LINKED ISSUE #263: Mark a type as allowing any value
Would it be possible to mark a bitflag as allowing any value?

This is useful to me when wrapping C apis that use bitflags where we as the user of the API may not be aware of all the possible flags. I would like to take the cint and turn it into a bitflags type, perserving the original value, while only defining the flag bits that I am aware of/are public/care about.

KodrAus:
Closes #168 
Closes #155 
Closes #218 
Closes #263 
Closes #228 

Following on from #264 

Part of #262 

This PR is a refactoring of the generated flags types to make the API ownership between us (`bitflags`) and them (callers of `bitflags!`) clearer. We generate a public type with an API that we can't make a lot of change to outside of the `BitFlags` trait, and a private type that can hold trait implementations that users may choose to bubble up publicly.

This makes it possible to pick different semantics for traits like `Ord`, and add optional support for external libraries like `Arbitrary`.

It comes with some breakage though:

- If you previously had a `#[derive(Serialize)]` impl, you'll need a `#[serde(transparent)]` attribute to ensure your format doesn't change.
- If you `#[derive(Debug)]`, you'll now get your flags wrapped in the name of your type (newtype formatting).
- The `.bits` field is now a `.bits()` method.

We'll need to give this a thorough test once it builds to make sure it's workable for all the places `bitflags` is currently used, but I think this is a promising future direction for the project. Kind of like the `lazy_static` crate, a lot of our old complexity goes away when we don't need to work around some old language limitations.


src/lib.rs
```diff
@@ -285,9 +285,61 @@ mod bitflags_trait;
...
+    }
 }
+```
+
+What they'd end up with looks something like this:
+
+```rust
+pub struct MyFlags(<MyFlags as PublicFlags>::InternalFlags);
+
+const _: () {
```
konsumlamm, line 320:
```suggestion
const _: () = {
```

src/lib.rs
```diff
@@ -382,77 +432,83 @@ macro_rules! bitflags {
             }
         }
 
+        const _: () = {
```
konsumlamm, line 435: What's the reason for putting this in a `const _: () = { ... }`?

KodrAus: It turns out `const _: () = { .. }` is a great container for code generated by macros because:

1. It's supported in any context where other items can be generated. You can write them in modules, functions, and the middle of other expressions.
2. It gives us a new scope to define items in. Any types we define in the body of this `const` [don't conflict with any outside of it](https://gist.github.com/027b2678e6cd7c6f3ceadb23d785f38e). Any types we define in the `const` are also unreachable from outside of it. That's why we've got a trait with an associated type that lets us peek inside. We can even wrap the contents of the `const` body in a `mod` to hide fields if we want.

KodrAus:
cc @konsumlamm @Arturoc What do you all think of this new internal library organization? Are you happy with this direction?

konsumlamm:
I like it! No complaints from me.

arturoc:
looks good to me

KodrAus:
Alrighty! I think the only thing remaining before we could consider stabilizing is adding an associated type for the iterator to the `BitFlags` trait, so that you can name it.

I'll do that in a follow-up, then look at publishing a pre-release build.

Issue #285, opened 2022-09-12
Consider making a release


yshui:
1.3.2 is released 2 years ago and there are valuable changes on `main` since then. When will the next version be released?

PR #286, opened 2022-10-06
expose bitflags iters using nameable types

KodrAus:
Unblocks #285 
Part of #262, #28 

This PR is the last piece of the puzzle for a `2.0` release of the library that includes a `BitFlags` trait. It exposes iterators for the names and values of set flags that can be referenced through the `BitFlags` trait, such as `<Flags as BitFlags>::Iter`. The `wrapped_iter` UI test has an example of how to refer to the iterators.


KodrAus:
I think we can simplify the machinery between the public and internal types to use less indirection, now that all the impl blocks live alongside one another. This would also have the benefit of simplifying the types that appear in the docs.

PR #291, opened 2022-11-20
Add missing "if" to contains doc-comment in traits.rs

rusty-snake:


KodrAus (approved):
Thanks @rusty-snake! Well spotted

PR #294, opened 2023-01-02
Forbid unsafe_code

### LINKED ISSUE #222: forbid unsafe_code in the library code?
Right now, bitflags contains no unsafe blocks. User code can `#![forbid(unsafe_code)]`. But it would be nice to have  `#![forbid(unsafe_code)]` here too, to add to the assurance that the library doesn't contain any user-facing unsafe code if the user also uses `#![forbid(unsafe_code)]`. Would this be reasonable for bitflags?

I think it would have to be `#![cfg_attr(not(test), forbid(unsafe_code))]` to allow for tests using unsafe, but same difference.

fintelia:
Fixes #222. Contrary to what is mentioned there, removing this annotation would not constitute a breaking change requiring a major version bump (at least according to normal Rust conventions). It is merely an assertion that the _current_ patch release contains no unsafe code.

Looking for `#![forbid(unsafe_code)]` is more reliable than trying to grep the crate source code for instances of "unsafe", since the latter could hypothetically be concealed by macros and such.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -340,6 +340,7 @@
 
 #![cfg_attr(not(test), no_std)]
 #![doc(html_root_url = "https://docs.rs/bitflags/2.0.0-rc.1")]
+#![forbid(unsafe_code)]
 
 #[doc(inline)]
 pub use traits::BitFlags;


KodrAus (approved):
Thanks @fintelia :+1: I think this is something I'd be comfortable doing now. My comment about major version bumps was less about semver compatibility and more about making a reasonable commitment to users who care about the presence of unsafe code in their dependencies. You certainly should expect unsafe code to appear or disappear in patches, but I see `#[forbid(unsafe_code)]` also as a design guideline to avoid any solutions that require unsafe code.

PR #296, opened 2023-01-26
serde: enable no-std support

nim65s:
Hi,

This PR enable no-std support on serde, ref. https://serde.rs/no-std.html

Otherwise, trying to compile a project with bitflags 2.0.0-rc.1 -F serde  on --target=thumbv7m-none-eabi currently hangs with a huge number of eg.
```
error[E0432]: unresolved imports `self::core::cmp`, `self::core::iter`, `self::core::mem`, `self::core::num`, `self::core::ptr`, `self::core::slice`, `self::core::str`
   --> /home/nim/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.152/src/lib.rs:167:26
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ A macro to generate structures which behave like bitflags.
 exclude = ["tests", ".github"]
 
 [dependencies]
-serde = { version = "1.0", optional = true, default-features = false, features = ["derive"] }
+serde = { version = "1.0", optional = true, default-features = false }
 core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core" }
 compiler_builtins = { version = "0.1.2", optional = true }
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ A macro to generate structures which behave like bitflags.
 exclude = ["tests", ".github"]
 
 [dependencies]
-serde = { version = "1.0", optional = true }
+serde = { version = "1.0", optional = true, default-features = false, features = ["derive"] }
```
KodrAus, line 22:
```suggestion
serde = { version = "1.0", optional = true, default-features = false }
```

KodrAus (commented):
Thanks @nim65s! I don't think we need the `derive` feature here since we only derive `serde` traits in tests.

Cargo.toml
nim65s: exact, thanks for the review !

KodrAus (approved):
Thanks @nim65s! This is a great catch :+1:

PR #297, opened 2023-02-02
Add a parser for flags formatted as bar-separated-values

KodrAus:
The last blocker to #262

This PR adds a parser for flags formatted using the default `Debug` impl, like:

```
A | B | C | 0x0e
```

It also updates the default `serde` implementations to use the text format and parser for human-readable formats.

I'll update this PR with more details once I've got the parser properly written and tested.


src/error.rs
```diff
@@ -0,0 +1,88 @@
+use core::fmt;
+
+#[derive(Debug)]
+pub struct ParseError(ParseErrorKind);
```
KodrAus, line 4: Missing docs on this type

src/internal.rs
```diff
@@ -76,18 +90,71 @@ macro_rules! __impl_internal_bitflags {
...
-                    f.write_str("empty")?;
+                    $crate::__private::core::write!(f, "{:#x}", <$T as $crate::__private::Bits>::EMPTY)?;
                 }
 
                 $crate::__private::core::fmt::Result::Ok(())
             }
         }
 
+        // The impl for `FromStr` should parse anything produced by `Display`
+        impl $crate::__private::core::str::FromStr for $InternalBitFlags {
```
KodrAus, line 115: This format should be documented somewhere with a grammar so users can generate formatted strings themselves and know they're going to parse.

examples/serde_compat.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    impl<'de> serde::Deserialize<'de> for Flags {
+        fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+            legacy_format::deserialize(deserializer)
+        }
+    }
+
+    pub mod legacy_format {
```
KodrAus, line 31: I think I will pull this out into a library to make it easier to consume. Users would then be free to inline it from there if they wanted to.

KodrAus:
One reasonable enhancement we could consider in the future is introducing a trait for parsing:

```rust
pub trait Parser {
    type Bits;

    fn name(&mut self, flag: &str) -> Result<Self::Bits, ParseError>;
    fn hex(&mut self, flag: &str) -> Result<Self::Bits, ParseError>;

    fn whitespace(&mut self, ws: &str) -> Result<(), ParseError> {
        Ok(())
    }

    fn separator(&mut self, separator: &str) -> Result<(), ParseError> {
        Ok(())
    }
}
```

We could then add a method to our generated internal type (not the public one) like:

```rust
pub fn parse(parser: &mut impl Parser<$T>) -> Result<Self, ParseError> { .. }
```

Then, end-users could write their own implementation of a parser that could ignore invalid flags, or treat them as case-insensitive.

KodrAus:
I've published a library that maintains the `serde` compatibility functions: https://github.com/KodrAus/bitflags-serde-legacy/

KodrAus:
Last thing I think we should do before merging is run some benchmarks, just to make sure that the cost of parsing doesn't increase too badly with the number of flags.

KodrAus:
Some very simple benchmarking suggests around 10ns parsing per flag on a flags type with 10 flags. There would be plenty of room to improve this, but it doesn't look like we've got a risk of exploding parsing time out.

PR #307, opened 2023-03-11
fix a nit in the readme

s1rius:


README.md
```diff
@@ -40,7 +40,7 @@ bitflags! {
         const A = 0b00000001;
         const B = 0b00000010;
         const C = 0b00000100;
-        const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
+        const ABC = Self::A.bits() | Self::B.bits() | Self::C.bits();
```
tiaanl, line 43: Why do a function call when the value is there in a variable?

KodrAus: In `2.x` it's no longer in a field called `bits`, the library generates a newtype so the quickest migration is to use the method call instead.

KodrAus (approved):
Thanks @s1rius! This looks good to me

PR #313, opened 2023-03-15
Make empty_flag() const.

tormeh:
Admittedly a low-effort contribution


KodrAus (approved):
Thanks @tormeh! This looks good to me

Issue #310, opened 2023-03-14
2.0.0 changelog appears incomplete


MarijnS95:
#### Context

I maintain a few crates that provide `bitflags` types in their public API, while the crates themselves don't really use these types heavily. As such I won't notice changed/missing APIs until these crates are published and used out in the wild.

Most importantly, besides listing commit/PR titles in a changelog, it should serve as a human readable "migration guide" explaining the outcome or expected action from consumer crates. Right now this appears to be lacking, forcing consumers to either blindly whack some `derives` on types or features on the crate dependency, or search through this repository and PRs to understand whether that's the right/intended thing to do in the first place.

#### Cannot derive `serde` on `bitflags` anymore

Some digging seems points to `bitflags` now being a newtype wrapper around an internal implementation, which requires the `serde` feature (disabled by default in `serde: enable no-std support by @nim65s in https://github.com/bitflags/bitflags/pull/296`) to be enabled to get this propagated. The entire effect of this newtype wrapper should have had a honorable mention in the changelog :slightly_smiling_face:

**EDIT: Even worse, https://github.com/bitflags/bitflags/pull/282 makes the following mention:**
> If you previously had a `#[derive(Serialize)]` impl, you'll need a `#[serde(transparent)]` attribute to ensure your format doesn't change.

Again, such kind of breakage is exactly the thing to describe in the changelog :sweat_smile: 

#### Changes to `serde` serialization

`2.0.0-rc.2` has a big warning that the serialization format changed: https://github.com/bitflags/bitflags/blob/main/CHANGELOG.md#200-rc2, yet this warning was not propagated to the main 2.0.0 changelog: https://github.com/bitflags/bitflags/releases/tag/2.0.0

#### `Debug`, `Copy`, `Clone`, `Hash`, `(Partial)Eq` etc not automatically derived anymore

Also appears to be explicitly asked in the issues and happening as a result of the newtype wrapper, but nothing explaining this in the changelog. Would be helpful if it contains a full list of all the trait implementations that have been "removed" going from `1.3.2` to `2.0.0`.

---

Regardless of all that, thanks for making all these changes and improvements! I'm sure they'll get appreciated, as long as such breaking effects get the attention they deserve in the changelog.

MarijnS95:
[...]

PR #316, opened 2023-03-16
Fix formatting of multi-bit flags with partial overlap

### LINKED ISSUE #315: Display missing extra bits for multi-bit flags
See: https://github.com/bitflags/bitflags/issues/310#issuecomment-1470122112

Given a flags type with two flags, `BIT = 0b0000_0001` and `MASK = 0b0001_1110`, formatting the value `3` should result in `BIT | 0x2`, but instead it gives `BIT`. That extra bit gets lost, so doesn't roundtrip. The problem seems to be in the generated `iter_names` method.

KodrAus:
Fixes #315

cc @mxk

I've added a repro for the bug reported here: https://github.com/bitflags/bitflags/issues/310#issuecomment-1470122112

It looks like we'll need to tweak the `iter_names` method to make sure we include any remaining bits when they overlap, but don't complete, a multi-bit flag.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -89,7 +89,8 @@ macro_rules! __impl_internal_bitflags {
 
                 // Iterate over the valid flags
                 let mut first = true;
-                for (name, _) in self.iter_names() {
+                let mut iter = self.iter_names();
+                for (name, _) in &mut iter {
                     if !first {
                         f.write_str(" | ")?;
                     }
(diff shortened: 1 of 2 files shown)


KodrAus:
Ok, it looks like the fix for this was fairly straightfoward: we were using the flags themselves to try mask out any bits that don't correspond to a valid flag, but in this example the extra bits do correspond to _part_ of a flag. So we'd mask them out without printing anything for them.

Using the iterator as the source of these bits ensures anything not printed by name will get printed as a hex number.

PR #326, opened 2023-03-29
Support zerocopy derives

### LINKED ISSUE #325: Support zerocopy
As per #310, for users of bitflags to continue being able to derive [`zerocopy`](https://crates.io/crates/zerocopy) traits on bitflags-generated types, it appears that there will need to be a cargo feature flag to make the internal bitflags types derive the relevant zerocopy traits.

qwandor:
Fixes #325.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -49,7 +49,7 @@ jobs:
       run: cargo install cargo-hack
 
     - name: Powerset
-      run: cargo hack test --feature-powerset --lib --optional-deps "std serde" --depth 3 --skip rustc-dep-of-std
+      run: cargo hack test --feature-powerset --lib --optional-deps "std serde zerocopy_0_6" --depth 3 --skip rustc-dep-of-std
 
     - name: Docs
       run: cargo doc --features example_generated
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ exclude = ["tests", ".github"]
 serde = { version = "1.0", optional = true, default-features = false }
 core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core" }
 compiler_builtins = { version = "0.1.2", optional = true }
+zerocopy = { version = "0.6.1", optional = true, default-features = false }
 
 [dev-dependencies]
 trybuild = "1.0"
@@ -35,6 +36,7 @@ serde_test = "1.0"
 std = []
 example_generated = []
 rustc-dep-of-std = ["core", "compiler_builtins"]
+zerocopy_0_6 = ["zerocopy"]
 
 [package.metadata.docs.rs]
 features = ["example_generated"]
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -14,6 +14,12 @@ macro_rules! __declare_internal_bitflags {
         $iter_vis:vis struct $Iter:ident;
         $iter_names_vis:vis struct $IterNames:ident;
     ) => {
+        #[cfg(feature = "zerocopy_0_6")]
+        use $crate::__private::zerocopy;
+        #[cfg(feature = "zerocopy_0_6")]
+        use ::core::marker::Sized;
+
+        #[cfg_attr(feature = "zerocopy_0_6", derive($crate::__private::zerocopy::AsBytes, $crate::__private::zerocopy::FromBytes))]
         #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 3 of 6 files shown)


KodrAus:
Thanks for the PR @qwandor!

Before we do this, I think we need to decide how we want to support unstable dependencies in this library. I've opened #329 to do that. Any feedback would be much appreciated!

src/internal.rs
```diff
@@ -14,6 +14,12 @@ macro_rules! __declare_internal_bitflags {
         $iter_vis:vis struct $Iter:ident;
         $iter_names_vis:vis struct $IterNames:ident;
     ) => {
+        #[cfg(feature = "zerocopy_0_6")]
+        use $crate::__private::zerocopy;
+        #[cfg(feature = "zerocopy_0_6")]
+        use ::core::marker::Sized;
+
+        #[cfg_attr(feature = "zerocopy_0_6", derive($crate::__private::zerocopy::AsBytes, $crate::__private::zerocopy::FromBytes))]
```
KodrAus, line 22: Unfortunately this won't actually work. This code is expanded in an end-user's library where there's no knowledge of whether or not `bitflags` has this feature enabled. For other libraries, we instead expand to a macro that calls back into `bitflags` and generates appropriate `impl` blocks.

Since this library requires using their custom derive I think we'll need a different strategy that accepts the struct definition as input and annotates it with its attributes. We'll need a second macro that also includes those necessary `use` statements.

If `zerocopy` `0.7` that's currently in-progress has a nicer experience for implementing its traits then I think it would be good to wait for it.

.github/workflows/rust.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       run: cargo install cargo-hack
 
     - name: Powerset
-      run: cargo hack test --feature-powerset --lib --optional-deps "std serde" --depth 3 --skip rustc-dep-of-std
+      run: cargo hack test --feature-powerset --lib --optional-deps "std serde zerocopy_0_6" --depth 3 --skip rustc-dep-of-std
```
KodrAus, line 52: After rebasing on `main` this shouldn't be necessary anymore. We automatically test all of these features.

Cargo.toml
```diff
@@ -23,6 +23,7 @@ exclude = ["tests", ".github"]
 serde = { version = "1.0", optional = true, default-features = false }
 core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core" }
 compiler_builtins = { version = "0.1.2", optional = true }
+zerocopy = { version = "0.6.1", optional = true, default-features = false }
```
KodrAus, line 26: I think we should do something like:

```toml
zerocopy_0_6 = { version = "0.6.1", optional = true, default-features = false, package = "zerocopy" }
```

here and remove the explicit feature. Otherwise users can depend on `zerocopy` directly, which will become confusing when a new version is released.

KodrAus:
@qwandor It looks like this is going to be a bit trickier than I first thought. I'm thinking of adding a `bitflags_attributes` crate that can conditionally expand to derives for libraries like `zerocopy` that require them based on `bitflags` features.

If you're happy, I'll work on that and roll `zerocopy` support in with it.

KodrAus:
With #348 merged I've added an example of how to integrate `zerocopy` with a generated flags type [here](https://github.com/bitflags/bitflags/blob/main/examples/custom_derive.rs). We don't need to add direct support for the library in `bitflags` for you to derive it anymore.

src/internal.rs
joshlf: Hey, came across this while discussing https://github.com/google/zerocopy/issues/287. Zerocopy 0.7 has been released now. Is the experience you're referring to the need to emit an impl for the function `only_derive_is_allowed_to_implement_this_trait` or something else? We've recently done some work to [make the docs more precise](https://github.com/google/zerocopy/pull/283) with regards to what needs to hold in order for a trait impl to be sound, so that might help you. Let me know if there's anything else we could add that would make this easier for you!

joshlf: @qwandor I believe I've figured out a way to support this: https://github.com/google/zerocopy/issues/1497#issue-2393919599

If you're still interested in zerocopy support and have some cycles to donate, let me know. It'd involve adding support for a `#[zerocopy(crate = ...)]` attribute (https://github.com/google/zerocopy/issues/11), which we currently don't have cycles for on our side.

qwandor: I think #348 works for everything I need, so no explicit zerocopy support is needed here.

PR #331, opened 2023-04-03
Ignore clippy lint in generated code

Jake-Shadle:
If a project enables https://rust-lang.github.io/rust-clippy/master/index.html the code generated by the macro will emit the warning, which for many projects will also fail CI.

### DIFF
--- a/src/external/serde_support.rs
+++ b/src/external/serde_support.rs
@@ -39,7 +39,7 @@ where
         {
             type Value = T;
 
-            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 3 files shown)


Jake-Shadle:
Also adds the lifetime specifier for several `Formatter<'_>`.

KodrAus (approved):
Thanks @Jake-Shadle :+1:

We should probably add Clippy to our own CI too to make sure we catch these things rather than end-users having to.

Jake-Shadle:
It's a good thing to have for sure! Though I think you could still run into trouble with lints that are accept by default like this particular case, in addition to behavior changes (new lints/default changes/etc) between Rust stable versions.

Issue #332, opened 2023-04-05
insert_if() and remove_if()


fpagliughi:
The `BitFlags::set()` function is nice. Very convenient. But it would be nice to have two similar functions that set or clear the bit(s), respectively if the condition is true, but _do nothing_ if the condition is false. Like:

```
/// Inserts the specified flags if the value is true; does nothing if the value is false.
fn insert_if(&mut self, other: Self, value: bool);

/// Removes the specified flags if the value is true; does nothing if the value is false.
fn remove_if(&mut self, other: Self, value: bool);
```

KodrAus:
Hi @fpagliughi, thanks for the suggestion! These do seem like useful utility methods, but I don't think they're broadly useful enough to generate automatically. You can freely implement them on your own flags types though:

```rust
bitflags::bitflags! {
    pub struct Flags: u8 {
        const A = 0b0000_0001;
        const B = 0b0000_0010;
        const C = 0b0000_0100;
    }
}

impl Flags {
    pub fn insert_if(&mut self, other: Self, value: bool) {
        if value {
            self.insert(other);
        }
    }

    pub fn remove_if(&mut self, other: Self, value: bool) {
        if value {
            self.remove(other);
        }
    }
}
```

KodrAus:
This isn't something we'll support in `bitflags` directly, but can be freely added to your types through regular impl blocks. I'll go ahead and close this one, but please feel free to open new issues for any other suggestions!

PR #333, opened 2023-04-05
Implemented insert_if() and remove_if()

fpagliughi:
Well that was easy enough...

Implements the feature request in #332 to add new functions to the `bitflags` trait for `insert_if()` and `remove_if()`.


KodrAus:
Thanks for the PR @fpagliughi. As mentioned in #332, I think these would be better implemented on your flags types directly rather than generated for all flags types.

PR #336, opened 2023-04-11
Add support for bytemuck

### LINKED ISSUE #311: Support bytemuck
From #310

Add a `bytemuck` Cargo feature that derives `bytemuck` traits on the generated inner flags type so end-users can then `#[derive(bytemuck::*)]` traits on their own flags types.

KodrAus:
Closes #311 

I've also added some docs on how to support external libraries here for future contributors. Since the `bytemuck` impls themselves are `unsafe`, I've added a note to where we generate our internal flags type warning that its ABI can't be changed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ exclude = ["tests", ".github"]
 [dependencies]
 serde = { version = "1.0", optional = true, default-features = false }
 arbitrary = { version = "1.0", optional = true }
+bytemuck = { version = "1.0", optional = true }
 core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core" }
 compiler_builtins = { version = "0.1.2", optional = true }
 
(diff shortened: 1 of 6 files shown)


KodrAus:
cc @MarijnS95 do these look right to you? I haven't used `bytemuck` myself.

KodrAus:
cc @Lokathor :wave:

I just wanted to make sure I've got these `Pod` and `Zeroable` impls right 🙂 In `bitflags`, we can automatically implement them on an internal type that looks like:

```rust
#[repr(transparent)]
struct FlagsField<T> {
    bits: T
}
```

where `T` may be a numeric type, like `i32`, `usize` etc. We may allow you to specify this type as something you've invented at some point in the future though. The type will _always_ be a `#[repr(transparent)]` struct with a single field though.

The impls I've come up with [look like this](https://github.com/bitflags/bitflags/pull/336/files#diff-9be611bd21b41b3a4b178a702a082490e281827719e6ea1499448524d7cbc3b9R228-R244). Does that look right to you?

Lokathor:
yep, `impl<T> Pod for ReprTransparentIntNewtype<T> where T: Pod` is basically exactly what I was expecting to see before I clicked, and then it's what I saw.

KodrAus:
Awesome, thanks for checking them out! 🙇 I'll get this rolling through now.

PR #340, opened 2023-04-17
add docs to describe the behavior of multi-bit flags

nicholasbishop:
This clears up possible confusion about how a flag that defines multiple bits behaves when calling `from_bits` or `from_bits_truncate`. Those functions do not treat multi-bit flags as a set of possible bits that can be set, but rather as a unit where the bits must exactly match one of the defined flags.


KodrAus (approved):
Thanks @nicholasbishop! This is a great clarification :bow:

PR #351, opened 2023-05-04
Support ejecting flags types from the bitflags macro

### LINKED ISSUE #348: Allow external impls of Bits and BitFlags
The `BitFlags` trait is currently sealed, and is only supported through the `bitflags!` macro. I think we should make this trait publicly implementable, and default most of its members. I spent some time hacking on this, and came up with this minimal implementation.

Given a flags type like:

```rust
bitflags! {
    pub struct MyFlags: u32 {
        const A = 0b0000_0001;
        const B = 0b0000_0010;
        const C = 0b0000_0100;
    }
};
```

You can implement it manually with:

```rust
pub struct MyFlags {
    bits: u32,
}

impl BitFlags for MyFlags {
    const FLAGS: &'static [(&'static str, Self::Bits)] = &[
        ("A", 0b0000_0001),
        ("B", 0b0000_0010),
        ("C", 0b0000_0100),
    ];

    type Bits = u32;
    type Iter = bitflags::iter::Iter<Self>;
    type IterNames = bitflags::iter::IterNames<Self>;

    fn bits(&self) -> Self::Bits {
        self.bits
    }

    fn from_bits_retain(bits: Self::Bits) -> Self {
        Self {
            bits
        }
    }

    fn iter(&self) -> Self::Iter {
        bitflags::iter::Iter::new(self)
    }

    fn iter_names(&self) -> Self::IterNames {
        bitflags::iter::IterNames::new(self)
    }
}
```

I'm proposing we _don't_ do #293, so that the implementation of `BitFlags` doesn't call for a host of supertraits.

The `FLAGS` constant there is new, and drives the iteration-based methods like `from_bits_truncate`, `from_name`, and the implementations of `Iter` and `IterNames`. If you squint, it looks a lot like the body of the `bitflags` macro.

I think doing this has a few benefits:

1. It makes the trait "real", so you can implement it and work with it yourself.
2. It moves most of the generated code out of macros where they don't need to use fully-qualified paths like `$crate::__private::core::option::Option::Some<T>`.
3. It gives us a starting point for #347 and #339, where you might not be able to use the `bitflags!` macro, but with a small amount of effort, can cook up your own flags type with whatever shape you want, and still get the benefits of generated code.

For trait impls like `serde` and `arbitrary`, we can then expose utilities like that proposed `bitflags::iter` module that make manual impls of those traits with awareness that the type is a flags enum easy:

```rust
impl Serialize for MyFlags {
    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        bitflags::serde::serialize(self, serializer)
    }
}
```

I'm keen to hear what people think of this.

KodrAus:
Closes #348 

This PR lets you define your flags type outside of the `bitflags!` macro, but still optionally use it to generate constants, methods, and trait impls for your flags types:

```rust
// First: Define your flags type. It needs to be a newtype over its underlying bits type
pub struct ManualFlags(u32);

// Next: use `impl Flags` instead of `struct Flags`
bitflags! {
    impl ManualFlags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
        const ABC = Self::A.bits() | Self::B.bits() | Self::C.bits();
    }
}
```

The `BitFlags` trait that can't be manually implemented has also been deprecated in favour of a semver-compatible `Flags` trait that can be manually implemented. This makes it possible to define flags types without using any macros at all:

```rust
// First: Define your flags type. It just needs to be `Sized + 'static`.
pub struct ManualFlags(u32);

// Not required: Define some constants for valid flags
impl ManualFlags {
    pub const A: ManualFlags = ManualFlags(0b00000001);
    pub const B: ManualFlags = ManualFlags(0b00000010);
    pub const C: ManualFlags = ManualFlags(0b00000100);
    pub const ABC: ManualFlags = ManualFlags(0b00000111);
}

// Next: Implement the `BitFlags` trait, specifying your set of valid flags
// and iterators
impl Flags for ManualFlags {
    const FLAGS: &'static [Flag<Self>] = &[
        Flag::new("A", Self::A),
        Flag::new("B", Self::B),
        Flag::new("C", Self::C),
    ];

    type Bits = u32;

    fn bits(&self) -> u32 {
        self.0
    }

    fn from_bits_retain(bits: u32) -> Self {
        Self(bits)
    }
}
```


KodrAus:
Since we can't yet replace our `const` functions with default impls from the `Flags` trait we currently have 2 copies of each method. These will need a really good test to make sure they're equivalent.

KodrAus:
Before merging this, we'll need to introduce a bound that ensures you can't use a custom implementation of `Bits` in types _defined_ through the `bitflags` macro. Otherwise crate features in `bitflags` may introduce additional trait bounds that aren't satisfied by those custom types.

KodrAus:
Just need to come up with a 1.56.0 compatible method of asserting a custom `Bits` type is a Rust primitive. That will probably just be a private trait instead of a function call.

src/external/arbitrary.rs
```diff
@@ -1,3 +1,16 @@
+//! Specialized fuzzing for flags types using `arbitrary`.
+
+use crate::Flags;
+
+pub fn arbitrary<'a, B: Flags>(
```
KodrAus, line 5: Missing docs here

src/lib.rs
```diff
@@ -619,8 +618,62 @@ macro_rules! bitflags {
...
+            dead_code,
+            deprecated,
+            unused_doc_comments,
+            unused_attributes,
+            unused_mut,
+            unused_imports,
+            non_upper_case_globals
+        )]
+        const _: () = {
+            // NOTE: We don't need to check `__is_primitive` here because we don't derive
```
KodrAus, line 663: Out-of-date: This is now a trait implemented by declaring the internal bits field.

src/parser.rs
```diff
@@ -28,7 +28,113 @@
...
 
-use core::fmt;
+use core::fmt::{self, Write, LowerHex};
+
+use crate::{Flags, Bits};
+
+/// Write a set of flags to a writer.
+///
+/// Any bits that don't correspond to a valid flag will be formatted
+/// as a hex number.
```
KodrAus, line 38: Note that the output is guaranteed to roundtrip through `from_str`.

src/parser.rs
```diff
@@ -28,7 +28,113 @@
...
+}
+
+/// Parse a set of flags from text.
+///
+/// This function will fail on unknown flags rather than ignore them.
+pub fn from_str<B: Flags>(input: &str) -> Result<B, ParseError>
+where
+    B::Bits: FromHex,
+{
+    let input = input.trim();
```
KodrAus, line 94: Is this initial `trim` necessary? We trim each flag anyways

src/parser.rs
```diff
@@ -28,7 +28,113 @@
...
+
+use crate::{Flags, Bits};
+
+/// Write a set of flags to a writer.
+///
+/// Any bits that don't correspond to a valid flag will be formatted
+/// as a hex number.
+pub fn to_writer<B: Flags>(flags: &B, mut writer: impl Write) -> Result<(), fmt::Error>
+where
+    B::Bits: LowerHex,
```
KodrAus, line 41: We have the implicit requirement that `LowerHex` will write the `0x` prefix.

PR #354, opened 2023-05-17
fix clippy assign_op_pattern warnings

warthog618:
I'm currently seeing clippy warnings like:
```
warning: manual implementation of an assign operation
   --> uapi/src/v1.rs:457:1
    |
457 | / bitflags! {
458 | |     /// Additional configuration flags for event requests.
459 | |     #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
460 | |     pub struct EventRequestFlags: u32 {
...   |
469 | |     }
470 | | }
    | |_^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assign_op_pattern
    = note: this warning originates in the macro `__impl_public_bitflags` which comes from the expansion of the macro `bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This patch addresses those by replacing occurrences of `a = a op b` to `a op= b`, as suggested by the info link.
Not sure of that is all of them - but it is at least all that my code was triggering.

### DIFF
--- a/src/public.rs
+++ b/src/public.rs
@@ -173,7 +173,7 @@ macro_rules! __impl_public_bitflags {
                             $(#[$attr $($args)*])*
                             {
                                 if bits & $PublicBitFlags::$Flag.bits() == $PublicBitFlags::$Flag.bits() {
-                                    truncated = truncated | $PublicBitFlags::$Flag.bits()
+                                    truncated |= $PublicBitFlags::$Flag.bits()
                                 }
                             }
                         );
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for the PR @warthog618! We actually use the non-assigning variants intentionally here so that the trait bounds required by the underlying bits types is limited to `BitOp` rather than `BitOp` plus `BitOpAssign`.

I think the fix here would be to allow that lint.

It's probably also really about time I add `clippy` to CI here so these lints stop triggering in end-users code 🙂

warthog618:
Fair enough.  I'd be happy with any solution that will keep clippy quiet, so whatever works for you.

PR #361, opened 2023-06-13
[doc] [src/lib.rs]  delete redundant path prefix

OccupyMars2025:


KodrAus:
Thanks @OccupyMars2025!

PR #366, opened 2023-06-26
Fix up incorrect sub assign behavior and other cleanups

### LINKED ISSUE #363: Treatment of unknown bits in operators is inconsistent
The [`complement` function](https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#method.complement)'s documentation says "the returned set contains all the flags which are not set in `self`, but which are allowed for this type". So primarily here, the phrase "allowed for this type" is ambiguous; the word "allowed" doesn't appear to be used anywhere else in the documentation, and bitflags in general seems to "allow" bits that aren't defined, with the new `from_bits_retain` function.

Looking at the code, it appears `complement` does a `from_bits_truncate`. Secondarily, I just converted some existing code to bitflags 2.0 and found this behavior surprising. Bitflags allows users to create values with `from_bits_retain`, and I assumed that applying `!` would act like a plain bitwise complement, and had a debugging adventure trying to figure why code that looked like it was setting a flag wasn't actually setting it.

### LINKED ISSUE #364: Inconsistent debug output for flag with no bits
In a bitflags type where one of the named value has the value 0, the debug output for the type sometimes includes that value by name, and sometimes doesn't, apparently depending on whether any unrecognized bits are present. For example, this:
```rust
use bitflags::bitflags;

bitflags! {
    #[derive(Debug)]
    pub struct Flags: u32 {
        const RDONLY = 0;
        const WRONLY = 1;
    }
}

fn main() {
    println!("{:?}", Flags::RDONLY);
    println!("{:?}", Flags::from_bits_retain(0x100));
}
```
prints
```
Flags(0x0)
Flags(RDONLY | 0x100)
```
I don't have an opinion about whether it should print `RDONLY` in both or neither, but printing it in just one is confusing.

### LINKED ISSUE #365: SWC - Update to 2.3.x causes incorrect output
Linking from this issue here -
https://github.com/swc-project/swc/issues/7513

There was a problem in SWC where commas were being omitted in the output. It was determined that this was caused by `bitflags`. Something changed between `2.2.1` to `2.3.x` which broke SWC's output.

(I'm not the maintainer of SWC, so I can't offer any input on details on SWC's usage of `bitflags`, or what/how/why anything works/doesn't work)

Anyways, as of this filing date, all users who start new SWC projects (because they aren't cargo locked to an older version of bitflags) will get broken output due to this.

Edit: I have a reproducible repo [here](https://github.com/MolotovCherry/swc-bug-7513). You can verify downgrading `bitflags` fixes it with `cargo update -p bitflags@2.3.1 --precise 2.2.1`

KodrAus:
Closes #365 
Closes #364 

This PR reworks our test suite to more rigorously test each method on flags types and fixes up the issue of `-` and `-=` having different results with unknown bits.

This PR makes the following behavioral changes and clarifications:

1. `-=` now behaves the same as `-` and `difference`: It respects set bits that don't correspond to known flags.
2. Zero-valued flags are never printed in formatting output.
3. Composite flags like `const ABC = A | B | C` won't be printed in formatting output if `A`, `B`, and `C` have already been printed. Formatting just prints _some_ set of flags that will produce an equal bitpattern when parsed. It doesn't necessarily print every flag that intersects the value.
4. `a.difference(b)` is not the same as `a & !b` when there are set bits that don't correspond to known flags. `a.difference` will respect them, but `!` will unset them.
5. `!` will use `!self.bits() & Self::all().bits()` instead of `Self::from_bits_truncate(!self.bits())`. This is a very subtle distinction. `Self::from_bits_truncate` would remove `0b0000_0001` if the only valid flag was `0b1111_1111`, but `& Self::all().bits()` will retain `0b0000_0001`. It retains more bits when there are bits that don't correspond to a known flag, and there is a multi-bit flag defined that does cover them. This is unlikely to have any practical implication until we recommend the flag `const ALL = !0`.


src/iter.rs
```diff
@@ -93,24 +93,26 @@ impl<B: 'static> IterNames<B> {
...
-            if self.state.is_empty() {
+            if self.remaining.is_empty() {
                 return None;
             }
 
             self.idx += 1;
 
             let bits = flag.value().bits();
 
+            // TODO: Try another state value; `|` the bits into it, and don't yield unless they change the value of the output; this would avoid composite flags
```
KodrAus, line 114: Don’t need this: we can just xor the incoming flag against our state and if it’s non-zero then it represents a “new” flag we haven’t yielded yet.

src/parser.rs
```diff
@@ -48,10 +48,15 @@ where
...
     // but don't correspond to any flags.
 
-    // Iterate over the valid flags
+    // Iterate over known flag values
     let mut first = true;
     let mut iter = flags.iter_names();
-    for (name, _) in &mut iter {
+    for (name, f) in &mut iter {
+        // Ignore any zero-valued flags
+        if f.is_empty() {
```
KodrAus, line 56: Should become unnecessary once we do the xor in `IterNames`.

src/tests/fmt.rs
```diff
@@ -0,0 +1,33 @@
+use super::*;
+
+#[test]
+fn cases() {
```
KodrAus, line 4: Add hex and binary cases here

KodrAus:
cc @sunfishcode

This cleans up the issue around `-=`, some formatting, and starts to document what operators and methods are bitwise. I'd like to spend some more time on docs to clarify the difference between a _flag_ and a _bit_. It seems like we're working on a bit of a shaky conceptual foundation and it would be good to tighten it all up.

I haven't changed the behavior of `!` to respect bits that don't correspond to known flags, because it seems like an easy way to accidentally pollute your flags with bits you'd never mean to set.

KodrAus:
In the interests of getting a fix out for the current regression around `-=`, I'll go ahead and merge this and release as `2.3.3`. I'll keep #363 open to talk about unifying operators with unknown bits.

PR #373, opened 2023-07-07
allow clippy::bad_bit_mask in generated Debug impl

ComputerDruid:
Nightly clippy seems to have learned to be able to see 0 values in fields of const structs. This is causing a bunch of warnings in bitflags usages which have consts defined as 0.

I only see this issue on the 1.0 branch.

Example reproducing the issue:

```rust
use bitflags::bitflags;

bitflags! {
    struct Foo : u32 {
        const BAR = 0;
    }
}

fn main() {
    let f = Foo::BAR;
    dbg!(f);
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -455,6 +455,7 @@ macro_rules! __impl_bitflags {
                     $(
                         __impl_bitflags! {
                             #[allow(deprecated)]
+                            #[allow(clippy::bad_bit_mask)]
                             #[inline]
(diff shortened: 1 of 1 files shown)


KodrAus:
Hi @ComputerDruid :wave:

Thanks for the PR! Unless there's some critical safety issue that requires patching I don't think we'll make any more releases of `1.x`. If possible, upgrading to `2.x` would be the way to go, which doesn't require any major version bumps in your own libraries.

PR #388, opened 2023-11-29
Add support for impl mode structs to be repr(packed)

GnomedDev:
This allows the following code to compile, with the only downside being a possible copy of the internal value for non-packed types, although I'm somewhat certain that would be optimised out.
```rs
#[repr(packed)]
struct Example(u64);

bitflags::bitflags! {
    impl Example: u64 {
        ...
    }
}
```

### DIFF
--- a/src/public.rs
+++ b/src/public.rs
@@ -321,7 +321,8 @@ macro_rules! __impl_public_bitflags_ops {
                 &self,
                 f: &mut $crate::__private::core::fmt::Formatter,
             ) -> $crate::__private::core::fmt::Result {
-                $crate::__private::core::fmt::Binary::fmt(&self.0, f)
+                let inner = self.0;
+                $crate::__private::core::fmt::Binary::fmt(&inner, f)
             }
         }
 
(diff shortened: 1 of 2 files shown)


GnomedDev:
Seems like CI failing is unrelated?

Nilstrieb:
This should probably have a test to ensure it continues to work.

GnomedDev:
Seems like there aren't any tests for `impl` other than the `examples`, should I add a file for specifically this in `src/tests`?

KodrAus:
Hi @GnomedDev :wave:

Thanks for the PR. If you add a test to `tests/compile-pass` called something like `bitflags_impl_repr_packed` with your example that should make sure we don't regress anything. The current UI failure is unrelated to your changes. If you rebase on `main` they should start working again.

GnomedDev:
Hey @KodrAus, just done that!

KodrAus (approved):
Thanks @GnomedDev! This looks good to me.

GnomedDev:
Thanks for merging, any chance for a release now with this in? I'm stoked to use it!

PR #392, opened 2024-01-16
Delete use of `local_inner_macros`

dtolnay:
`macro_export(local_inner_macros)` is an outdated workaround that is only relevant for supporting compilers from Rust 1.29 and older.

Currently bitflags's oldest supported compiler is Rust 1.56, so there should be no reason to use `local_inner_macros`.


dtolnay:
CI failure is unrelated to this change. I put up a fix in https://github.com/bitflags/bitflags/pull/393.

PR #398, opened 2024-02-06
Derive `Debug` for `Flag<B>`

tgross35:
This just makes it possible to easily print the `FLAGS` constant.


KodrAus (approved):
Thanks @tgross35! This looks good to me.

PR #402, opened 2024-03-20
Sync CHANGELOG.md with github release notes

dextero:
I was updating another crate to use bitflags v2 by using the changelog file as a sort of migration guide. Only when I was done going through the list of PRs I learned that the github release notes contained information than would have been useful :sweat_smile: 

So I pulled all the release notes as described on https://stackoverflow.com/a/76996206, and copied over any significant information I found missing to the CHANGELOG.md.

This might be helpful to anyone else like me who doesn't immediately think about looking at release notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,20 @@ to a valid flag.
 
 # 2.3.0
 
+## Major changes
+
+### `BitFlags` trait deprecated in favor of `Flags` trait
+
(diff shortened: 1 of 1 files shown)


KodrAus (approved):
Thanks @dextero! I should probably adjust my release workflow here a bit. I usually pull the diff straight into the CHANGELOG when assembling a release PR, and then come up with any more descriptive announcement when publishing the actual release. Going forwards I'll try write the full content of the release into the CHANGELOG and then populate the GitHub release from that.

KodrAus:
The build failure here is unrelated; I'll fix that up separately.

PR #404, opened 2024-03-21
Bump minimum declared versions of dependencies

dextero:
When updating `bitflags` dependency of another crate, I found out that `bitflags`' declared dependency versions are lower than needed to actually build the crate.

This PR adds a GH workflow check inspired by the one from [`gbm.rs`](https://github.com/Smithay/gbm.rs/blob/754a9574bc879bb1256ab058ded0c4fa598dcbf6/.github/workflows/ci.yml#L95) that attempts a build with the minimal compatible versions of all dependencies.

Another commit in the PR bumps the declared versions to earliest ones I identified as working, to make the workflow pass.

The workflow builds all features other than `core` and `rustc-dep-of-std`. `cargo check --all --all-features --all-targets` failed for me on main with mentions of rustc/std, which successfully scared me away :laughing: I have zero experience hacking on rustc/std so I don't even know how those are used.

Hopefully this, or some parts of it can be useful as is :smile:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,6 +38,20 @@ jobs:
     - name: Smoke test
       run: cargo run --manifest-path tests/smoke-test/Cargo.toml
 
+  check-minimal:
+    name: Check minimal versions
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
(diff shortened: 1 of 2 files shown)


KodrAus (approved):
Thanks @dextero! This is a great improvement. The updates to CI are much appreciated.

PR #414, opened 2024-07-03
Fix `clippy::doc_lazy_continuation` lints

waywardmonkeys:


KodrAus:
Thanks @waywardmonkeys! The build failure here will be unrelated; I'll look into that.

PR #416, opened 2024-07-26
Use `serde_derive` via `derive` feature on `serde`

waywardmonkeys:
This is the recommended way to do this in the `serde` docs.


waywardmonkeys:
The CI issue is not from this patch, so I've submitted #417 to start addressing it.

waywardmonkeys:
This is for the dev dependency only, so won't impact dependent crates of `bitflags`, so it won't run afoul of https://github.com/serde-rs/serde/issues/2584

PR #417, opened 2024-07-26
Fix CI: trybuild refresh, allow some clippy restrictions.

waywardmonkeys:
This updates the `trybuild` stuff to match the current compiler.

It also updates the clippy configuration within the smoke test to pass. If the smoke tests are only checked with `beta` channel clippy, perhaps we could do differently. Should we? If so, as part of this PR or separate?


waywardmonkeys:
I split this into 2 separate commits, so each one is fixing one part of the problem.

KodrAus (approved):
Thanks @waywardmonkeys!

Issue #375, opened 2023-07-24
Match expression question


juliusl:
How come the following isn't valid? 

```
bitflags! {
    #[derive(PartialEq, Eq, PartialOrd, Ord)]
    pub struct U32: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
    }
}

#[test]
fn test_flags() {
    match U32::A | U32::B {
        U32::A | U32::B => {},
        _ => panic!()
    }
}
```

juliusl:
Oh wait, I think I understand, this conflicts with the match `|` semantics.

kpreid:
This code would compile:
```rust
    const AB: U32 = U32::A.union(U32::B);
    match U32::A | U32::B {
        AB => {},
        _ => panic!()
    }
```
but the `match` will always perform an exact equality comparison, not a bit-test of any kind.

Issue #201, opened 2019-11-25
matching on bitflags leads to unreachable patterns


gz:
I was surprised to find out that matching on bitflags leads to unreachable patterns in the example below (i.e. `match_flag` always return 3). I'm assuming it conflicts with the `|` syntax in match arms, is this the intended behavior of bitflags?

```
use bitflags::bitflags;

bitflags! {
    pub struct Flag: u64 {
        const A      = 0;
        const B      = 1;
        const C      = 2;
    }
}

fn match_flag(f: Flag) -> usize {
    let a = Flag::A;
    let ab = Flag::A | Flag::B;
    let abc = Flag::A | Flag::B | Flag::C;
    match f {
        abc => 3,
        ab => 2,
        a => 1,
    }
}

fn main() {
    let abc = Flag::A | Flag::B | Flag::C;
    let ab = Flag::A | Flag::B;
    let a = Flag::A;
    
    assert_eq!(match_flag(abc), 3);
    assert_eq!(match_flag(ab), 2);
    assert_eq!(match_flag(a), 2);
}
```

Alexhuszagh:
@gz I don't believe you know how Rust matches work.

Try the above, except with the following `match_flag` and you get the exact same issue:

```
fn match_flag(f: Flag) -> usize {
    match f {
        abc => 3,
        ab => 2,
        a => 1,
    }
}
```

What does that mean? It means the matches aren't actually considering the left-hand value, it's just used as a name for binding.

Alexhuszagh:
@gz You can match the value of declared constants on the left-hand side, however, this is a bit tricky since bitflags does not override `|` as a a const fn.

What you can do however is:

```rust
fn match_flag(f: Flag) -> usize {
    const A: Flag = Flag::A;
    const AB: Flag = Flag::from_bits_truncate(Flag::A.bits() | Flag::B.bits());
    const ABC: Flag = Flag::from_bits_truncate(Flag::A.bits() | Flag::B.bits() | Flag::C.bits());
    match f {
        ABC => 3,
        AB => 2,
        A => 1,
        _ => 4,   // Needed for fallthrough
    }
}
```

Alexhuszagh:
However, you will notice this **still** won't give you what you want, because `bitflags` and `0` work weirdly together: in short, `Flag::A` will match everything, because that's how 0 works: `Flag::A & flag == Flag::A` by default.

If you change your full example to:

```rust
extern crate bitflags; // 1.2.1

[...]

PR #423, opened 2024-10-24
feat(core): Add match_bitflag macro for bitflag matching

YuniqueUnic:
I encountered an “unreachable” issue while utilizing Rust’s match statement. In search of a more effective solution, I considered employing a macro. However, I found that there wasn’t an existing macro to address this issue. Consequently, I decided to create one myself. 

Could you please review my code? I’m looking for feedback and suggestions. If you find it beneficial and suitable, I would appreciate it if you could merge it. Thank you.

related issues: 
1. #201 
2. #375 
3. ...

--- 

Implemented a new macro, `match_bitflag`, to perform matching operations on bitflags, similar to Rust's match expression. This macro addresses the matching issues that arise when using regular match expressions with bitflags.

- The macro supports complex bitflag combination matching.
- Added unit tests to verify the correctness of the macro.
- Provided usage examples and an explanation of the macro's internal implementation.

```rust
/// Match bitflag patterns similar to Rust's match expression
///
/// This macro allows for matching bitflag combinations in a way that resembles
/// Rust's native match expression, but works correctly with bitflag operations.
///
/// # Requirements
///
/// - The struct used with this macro must implement `PartialEq`.
/// - It's recommended to use this macro with structs created by the `bitflags!` macro.
///
/// # Syntax
///
/// ```ignore
/// match_bitflag!(expression, {
///     pattern1 => result1,
///     pattern2 => result2,
///     ...
///     _ => default_result
/// })
/// ```
///
/// # Examples
///
/// ```rust
/// use bitflags::{match_bitflag, bitflags};
///
/// bitflags! {
///     #[derive(PartialEq)]
///     struct Flags: u8 {
///         const A = 1 << 0;
///         const B = 1 << 1;
///         const C = 1 << 2;
///     }
/// }
///
/// let flags = Flags::A | Flags::B;
///
/// match_bitflag!(expected, {
///     Flags::A => println!("A"),
///     Flags::B => println!("B"),
///     Flags::C  => println!("C"),
///     Flags::A | Flags::B =>{
///          // println!("A | B");
///          print!("A");
///          print!(" | ");
///          print!("B");
///     },
///     Flags::A | Flags::C => println!("A | C"),
///     Flags::B | Flags::C => println!("B | C"),
///     Flags::A | Flags::B | Flags::C => println!("A | B | C"),
///     _ => println!("other")
/// })
/// ```
///
/// # How it works
///
/// The macro expands to a series of if-else statements, checking equality
/// between the input expression and each pattern. This allows for correct
/// matching of bitflag combinations, which is not possible with a regular
/// match expression due to the way bitflags are implemented.
///
/// # Note
///
/// The order of patterns matters. The first matching pattern will be executed,
/// so more specific patterns should come before more general ones.
#[macro_export]
macro_rules! match_bitflag {
    ($operation:expr, {
    $( $pattern:expr => $result:expr, )*
        _ => $default:expr
    }) => {
        {
            $(
                if $operation == $pattern {
                    $result
                } else
            )*
            {
                $default
            }
        }
    };
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -587,6 +587,90 @@ macro_rules! bitflags {
     () => {};
 }
 
+/// Match bitflag patterns similar to Rust's match expression
+///
+/// This macro allows for matching bitflag combinations in a way that resembles
+/// Rust's native match expression, but works correctly with bitflag operations.
+///
+/// # Requirements
+///
+/// - The struct used with this macro must implement `PartialEq`.
+/// - It's recommended to use this macro with structs created by the `bitflags!` macro.
+///
+/// # Syntax
+///
+/// ```ignore
+/// match_bitflag!(expression, {
+///     pattern1 => result1,
+///     pattern2 => result2,
+///     ...
+///     _ => default_result
+/// })
+/// ```
+///
+/// # Examples
+///
+/// ```rust
+/// use bitflags::{match_bitflag, bitflags};
+///
+/// bitflags! {
+///     #[derive(PartialEq)]
+///     struct Flags: u8 {
+///         const A = 1 << 0;
+///         const B = 1 << 1;
+///         const C = 1 << 2;
+///     }
+/// }
+///
+/// let flags = Flags::A | Flags::B;
+///
+/// match_bitflag!(expected, {
+///     Flags::A => println!("A"),
+///     Flags::B => println!("B"),
+///     Flags::C  => println!("C"),
+///     Flags::A | Flags::B =>{
+///          // println!("A | B");
+///          print!("A");
+///          print!(" | ");
+///          print!("B");
+///     },
+///     Flags::A | Flags::C => println!("A | C"),
+///     Flags::B | Flags::C => println!("B | C"),
+///     Flags::A | Flags::B | Flags::C => println!("A | B | C"),
+///     _ => println!("other")
+/// })
+/// ```
+///
+/// # How it works
+///
+/// The macro expands to a series of if-else statements, checking equality
+/// between the input expression and each pattern. This allows for correct
+/// matching of bitflag combinations, which is not possible with a regular
+/// match expression due to the way bitflags are implemented.
+///
+/// # Note
+///
+/// The order of patterns matters. The first matching pattern will be executed,
+/// so more specific patterns should come before more general ones.
+#[macro_export]
+macro_rules! match_bitflag {
+    ($operation:expr, {
+    $( $pattern:expr => $result:expr, )*
+        _ => $default:expr
+    }) => {
+        {
+            $(
+                if $operation == $pattern {
+                    $result
+                } else
+            )*
+            {
+                $default
+            }
+        }
+    };
+}
+
 /// Implement functions on bitflags types.
 ///
 /// We need to be careful about adding new methods and trait implementations here because they
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -18,6 +18,7 @@ mod intersects;
 mod is_all;
 mod is_empty;
 mod iter;
+mod match_bitflag;
 mod parser;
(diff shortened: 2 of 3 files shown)


KodrAus:
Hi @YuniqueUnic :wave:

It's a bit of a papercut that `A | B` becomes an or-pattern inside a match expression instead of an equality check on a bitor'd value for flags. I think enough of one that having a macro for matching is worth it.

Would you be happy to rename this to `bitflags_match!` and follow the same syntax as match arms do? Match arms don't require a trailing `,` if the body of the match is a block like:

```rust
match x {
    y => {} // <-- this is a block, so no trailing comma needed here
    z => (), // <-- this is not a block, so trailing comma needed here
}
```

YuniqueUnic:
Hello, @KodrAus 

Thanks for the feedback! I've made the changes:  

- Renamed the macro to `bitflags_match!` as suggested.  
- Improved the documentation and examples to reflect the changes.  
- Adjusted the test module names to match the new macro name. 

I tried to make the syntax follow the match arm rules (blocks do not require a trailing comma, non-blocks do), but it failed to compile, so for now, each pattern in the match still requires a trailing comma... I'm unable to solve this problem...

Please let me know if there are any other adjustments needed!

KodrAus:
Thanks @YuniqueUnic! I think we'll need to make trailing commas work correctly before we can ship this, but it may require some macro-hackery to make it work. I'd be happy to try take a look at it and see if I can come up with a solution.

KodrAus:
I've had a look at this and come up with a macro that I think matches most of the standard `match` syntax. You can see the diff here: https://github.com/KodrAus/bitflags/commit/925d014c56ec14f6873fd02fd37eab74f289b039, or the full macro here: https://github.com/KodrAus/bitflags/blob/925d014c56ec14f6873fd02fd37eab74f289b039/src/lib.rs#L657-L703

What I've written is a style of macro called a [tt-muncher](https://veykril.github.io/tlborm/decl-macros/patterns/tt-muncher.html). It lets us incrementally pull tokens and expand them as we go. The main benefit it gives us is a way to apply multiple rules to a collection of inputs. So we can have a rule for match arms where the body is surrounded by `{ }`, and another for match arms where the body isn't.

You should be able to pull this commit or my https://github.com/KodrAus/bitflags/tree/feat/bitflags_match branch into this PR.

YuniqueUnic:
@KodrAus 

Thank you for your support!!!

I forked your code, modified the doc of `bitflags_match` and I think I can learn the macro from you.
Now, I pulled the request again. 

Thanks!!

- Thanks to the help from @KodrAus, the syntax of `bitflags_match` is aligned with Rust's `match` syntax.
- Optimized the internal implementation of the macro so that it expands to a series of `if(pattern){ return result; }` statements.
- Updated the documentation to clarify the usage and precautions of the macro.
- Updated the test to verify that the macro is functioning correctly.

KodrAus (approved):
Thanks @YuniqueUnic! This looks good to me.

reneleonhardt:
@KodrAus So a logical contains one-of check still isn't possible I guess? 😞

```rust
// Check if flags contains B or C
let flags = Flags::A | Flags::B;
let check = Flags::B | Flags::C;
let b_or_c1 = bitflags_match!(flags, {
     check => true, // bitwise-or doesn't match logical B or C, only B + C combined
     _ => false,
});
let b_or_c2 = flags.contains(check); // false (same like bitflags_match, B + C combined)
let b_or_c3 = flags.contains(Flags::B) || flags.contains(Flags::C); // true (one of B or C)
let b_or_c4 = flags.one_of([Flags::B, Flags::C]); // one-of slice function doesn't exist
```

bitflags_match arms also don't allow complement check `b if !b.contains(A) => true,`.

Issue #425, opened 2024-12-03
Method to check if any unknown bits are set


atlv24:
When we deserialize bitflags from a different version of our software, we want to detect if there are bits we don't know about, and provide an Err Result describing as such. A `contains_unknown_bits` function would be great.

KodrAus:
Thanks for the suggestion @atlv24!

Right now, you can verify this with `let flags = Flags::from_bits(flags.bits()).ok_or_else(|| Error::new("flags type contains an invalid value"))?`, or with `flags & Flags::all() == flags`.

I'd also be in favor of adding a method to the `Flags` trait that told you if any unknown bits were set, along with a `truncate(&mut self)` method that unset them.

KodrAus:
This is done as of #426 and #428, Thanks @wysiwys!

PR #426, opened 2024-12-20
Add method to check if unknown bits are set

wysiwys:
This PR adds a method to the `Flags` trait, `contains_unknown_bits(&self)`, to check whether any unknown bits are set (See #425). It also adds tests for this method.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -22,6 +22,7 @@ mod parser;
 mod remove;
 mod symmetric_difference;
 mod union;
+mod unknown;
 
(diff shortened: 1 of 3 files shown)


KodrAus (approved):
Thanks @wysiwys! This looks great :+1:

KodrAus:
Build failure is unrelated; a few error messages have changed

cwfitzgerald:
Thank you for landing this!

Is there any reason these weren't added as inherent methods too?

PR #428, opened 2025-01-07
Add `truncate(&mut self)` method to unset unknown bits

wysiwys:
Hi,

This PR adds the method `truncate(&mut self)` to the `Flags` trait (see #425), which unsets any unknown bits. It also adds some tests for this method.


KodrAus (approved):
Thanks @wysiwys!

PR #437, opened 2025-02-28
`Flags` trait: add `clear(&mut self)` method

### LINKED ISSUE #418: No `clear()` method?
~~The previous major version used to have a `clear` method (i think?), while migrating to the new version i saw that this method was not present anymore, is this expected? it feels like a common behaviour to include, currently i'm doing this~~
I used to clear the flags by doing `self.bits = 0` in the old version of the library, in the current major version i saw this seems to be a way to do it instead, is there a reason why there is not a `.clear()` method or similar that unsets all bits?
```rust
bitflags! {
    #[wasm_bindgen]
    #[derive(Serialize, Copy, Clone, Debug)]
    pub struct Flags: u16 {
        const Carry    = 1<<1;
        const Overflow = 1<<2;
        const Zero     = 1<<3;
        const Negative = 1<<4;
        const Extend   = 1<<5;
    }
}

impl Flags {
    pub fn clear(&mut self) {
        self.set(Flags::all(), false);
        //or
       *self = Flags::empty();
    }
}
```

wysiwys:
This pull request adds a method to the `Flags` trait that clears all set bits in the flags. Resolves #418.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,6 +1,7 @@
 mod all;
 mod bitflags_match;
 mod bits;
+mod clear;
 mod complement;
 mod contains;
 mod difference;
--- a/src/tests/clear.rs
+++ b/src/tests/clear.rs
@@ -1,0 +1,27 @@
+use super::*;
+
+use crate::Flags;
+
+#[test]
+fn cases() {
+    case(TestFlags::from_bits_retain(0));
+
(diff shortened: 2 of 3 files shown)


KodrAus (approved):
Thanks @wysiwys! This looks good to me

KodrAus:
Build failure is because error messages have changed. I’ll fix that up in a followup PR

PR #462, opened 2025-09-21
Use serde_core instead of serde

james7132:
Serde recently split out the trait definitions from the top level serde crate. This crate does not rely on the derive macro, so it should be effective to depend on serde_core instead of serde. This doesn't impact the compilation times of glam itself, but allows it to parallelize with `serde-derive` and potentially other crates like `serde-json` that also do not rely on the derive macro, unblocking other crates reliant on bitflags to do so as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,14 +18,14 @@ A macro to generate structures which behave like bitflags.
 exclude = ["/tests", "/.github"]
 
 [dependencies]
-serde = { version = "1.0.103", optional = true, default-features = false }
+serde_core = { version = "1.0.103", optional = true, default-features = false }
 arbitrary = { version = "1.0", optional = true }
 bytemuck = { version = "1.12", optional = true }
 
 [dev-dependencies]
 trybuild = "1.0.18"
 rustversion = "1.0"
-serde_derive = "1.0.103"
+serde = { version = "1.0.103", features = ["derive"] }
 serde_json = "1.0"
 serde_test = "1.0.19"
 zerocopy = { version = "0.8", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


src/external.rs
```diff
@@ -155,7 +155,7 @@ macro_rules! __impl_external_bitflags_serde {
             ) -> $crate::__private::core::result::Result<Self, D::Error> {
                 let flags: $PublicBitFlags = $crate::serde::deserialize(deserializer)?;
 
-                Ok(flags.0)
+                $crate::__private::core::result::Result::Ok(flags.0)
```
KodrAus, line 158: Well spotted! :+1:

KodrAus (commented):
Thanks for working on this @james7132! I think we just need to update the `serde_core` version to one that's actually published, but otherwise this looks good to me.

Cargo.toml
```diff
@@ -35,6 +35,7 @@ bytemuck = { version = "1.12.2", features = ["derive"] }
 [features]
 std = []
 example_generated = []
+serde = ["serde_core"]
```
KodrAus, line 38: It looks like our MSRV doesn't like this, so we might need to use an alternative:

```toml
[dependencies]
serde = { version = "1.0.220", optional = true, default-features = false, package = "serde_core" }
```

src/external.rs
```diff
@@ -67,7 +67,7 @@ __impl_external_bitflags_my_library! {
 
 pub(crate) mod __private {
     #[cfg(feature = "serde")]
-    pub use serde;
+    pub use serde as serde;
```
KodrAus, line 70: This alias shouldn’t be necessary

KodrAus (commented):
Just spotted one more nitpick, then this should be good to go!

james7132:
Currently on vacation, will do so sometime next week.

PR #477, opened 2026-03-09
example_generated.rs: add missing third slash for doc comment

DanielEScherzer:


KodrAus (approved):
Thanks @DanielEScherzer!

DanielEScherzer:
`cargo test fail` locally
* works fine on rust 1.94.0 with my patch
* fails on beta with my patch - 1 failed, 1 filtered out
* fails on beta without my patch (current main) - 1 failed, 1 filtered out

So I could re-bless the tests, but that might make more sense to do as a separate patch - sent #478

PR #483, opened 2026-05-29
Add a custom `#[flag_name]` attribute

### LINKED ISSUE #470: Add ability to rename flags.
Various functions allow interacting with the names of flags as strings ([`Flag::name`](https://docs.rs/bitflags/latest/bitflags/struct.Flag.html#method.name), [`parser`](https://docs.rs/bitflags/latest/bitflags/parser/index.html), etc). However, these currently must work with the `SCREAMING_SNAKE_CASE` name of the flag as it appears in the source code, which is not always desirable.

In [Arti](https://arti.torproject.org), we use bitflags for our [router status flags](https://spec.torproject.org/dir-spec/consensus-formats.html#item:s), which have specific names assigned to them that do not match the Rust identifiers. It would be convenient to have a way of renaming flags with a attribute macro — [here's where this usecase came up](https://gitlab.torproject.org/tpo/core/arti/-/merge_requests/3454/diffs#diff-content-4c8b4a5b821efe6b52d4f98dfd87303c58db3166).

KodrAus:
Closes #470 

This PR introduces a `#[flag_name]` attribute you can apply to flags to give them a different name. I've tried to avoid adding these kinds of attributes in the past, but there's no way to get around this one with `bitflags`, so we support it. It lets you write:

```rust
bitflags! {
    pub struct MyFlags: u8 {
        #[flag_name = "a"]
        const A = 1;
    }
}
```

There isn't any validation done on the `flag_name` given, so you could pick strange names like `_`, or `a | b` or `  a`, etc. I need to add some more test coverage before this is ready.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -860,7 +860,7 @@ macro_rules! __bitflags_match {
     }
 }
 
-/// A macro that processed the input to `bitflags!` and shuffles attributes around
+/// A macro that processes the input to `bitflags!` and shuffles attributes around
 /// based on whether or not they're "expression-safe".
 ///
 /// This macro is a token-tree muncher that works on 2 levels:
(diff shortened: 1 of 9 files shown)


KodrAus:
Should add some compile-fail tests for malformed `#[flag_name]`s

PR #489, opened 2026-06-03
add MyFlags::Abc::iter_equal_names() method

ssrlive:


src/lib.rs
```diff
@@ -619,6 +623,24 @@ macro_rules! bitflags {
...
+#[doc(hidden)]
+macro_rules! __impl_public_bitflags_iter_equal_names {
+    ($BitFlags:ident) => {
+        impl $BitFlags {
+            /// Get an iterator over all defined names for this flags value.
+            ///
+            /// This iterator will yield all defined names for the flags value, including
+            /// any convenience flags.
+            #[inline]
+            pub fn iter_equal_names(&self) -> $crate::iter::ValueEqualNames<$BitFlags> {
```
KodrAus: We unfortunately can't add this as an inherent method on the generated flags type, because it can conflict with an inherent method already added by the user.

We'll need to add it to the `Flags` trait instead.

src/iter.rs
```diff
@@ -180,3 +180,35 @@ impl<B: Flags> Iterator for IterDefinedNames<B> {
         None
     }
 }
+
+/**
+An iterator over all defined names for a specific flags value.
+*/
+pub struct ValueEqualNames<T: Flags> {
```
KodrAus: `IterEqualNames`?

ssrlive:
I have changed the code following your suggestions.

src/iter.rs
```diff
@@ -180,3 +180,35 @@ impl<B: Flags> Iterator for IterDefinedNames<B> {
...
+An iterator over all defined names for a specific flags value.
+*/
+pub struct IterEqualNames<T: Flags> {
+    inner: IterDefinedNames<T>,
+    bits: T::Bits,
+}
+
+impl<T: Flags> IterEqualNames<T> {
+    #[inline]
+    pub fn new(value: &T) -> Self {
```
KodrAus: If possible we should make this `pub(crate)` to avoid exposing any more surface area than we need to.

ssrlive: fixed

KodrAus (approved):
Thanks @ssrlive! This looks good to me

PR #494, opened 2026-07-15
Flags: adjust order of elements in manual implementation example

DanielEScherzer:
Match the order given in the trait definition so that it is easier to compare the example with the requirements.


KodrAus (approved):
Thanks @DanielEScherzer!

PR #496, opened 2026-09-10
Re-pull const declarations outside of nested const

### LINKED ISSUE #495: ICE on Rust 1.69.0
We previously worked around a `rustdoc` bug: https://github.com/bitflags/bitflags/issues/320

But I removed that workaround while doing some other work on the macro internals. It seems like that's triggering an ICE on `1.69.0` for downstream crates: https://github.com/nix-rust/nix/pull/2812

Since our MSRV is advertised as 1.56.0, we should try to re-fix this issue.

cc @xtqqczze

KodrAus:
Fixes #495 

The const flag declarations were originally pulled out of a nested const block to work around a `rustdoc` bug. It seems like that same fix also prevented a `rustc` ICE on 1.69.0. I've re-pulled them back out.

@xtqqczze Do you still get an ICE triggered with this branch?


xtqqczze:
Confirmed this fixes the ICE locally. <strike>Let's wait for CI on https://github.com/nix-rust/nix/pull/2821 to confirm.</strike>

xtqqczze:
@KodrAus Confirmed that this fixes https://github.com/bitflags/bitflags/issues/495

KodrAus:
Thanks! I'll merge and publish

PR #499, opened 2026-09-16
Fix repeated input evaluation in bitflags_match!

christophacham:
Evaluate the input once without moving it. Add regression tests for later-arm matches, fallback matches, default-only invocations, and non-Copy inputs. Verified with cargo test --all-features and cargo fmt --all -- --check.

`bitflags_match!` could evaluate its input repeatedly—once per attempted match—instead of once overall.

So `bitflags_match!(next_flags(), …)` could call `next_flags()` several times and match against different values. With only a default arm, it never called it.

The fix evaluates the input once and reuses it without moving it. Four regression tests cover this.


KodrAus (approved):
Thanks @christophacham!

